PR #18, opened 2012-12-04
Fix HER-2023: ARCRecord.readHttpHeader() fails if HTTP response lacks empty line between HTTP headers and HTTP body.

aaronbinns:
The fix is to remove the unnecessary call to HttpParser.parseHeaders().  AFAICT, the parsed headers are never used (?).  We still want the HTTP _status_ line parsed of course, but not the rest of the headers.


ato:
This class is no longer part of the heritrix3 repo and the code in commons has changed quite a bit so I don't think there's any pint keeping this PR open.

PR #113, opened 2014-12-09
Allow non-login GET forms to be submitted if ExtractorHTMLForms.extractAllForms == true

vonrosen:
This is a preliminary attempt at allowing form submissions for non-login forms that have method="get". It is not complete and should NOT be merged. It is for review only.

The goal of this enhancement is to allow forms to be submitted if the do not seem like login forms, which are already handled by FormLoginProcessor, and the extractAllForms value on ExtractorHTMLForms is true. Each submit input found in the form creates a a new submission attempt. Dropdowns (<select>...</select>) are also parsed and their selected values are extracted. If there is no "selected" option value, then the first option value is chosen as the one to submit.

Before continuing on this enhancement, I want review input to determine if the solution looks desirable. There will need to be a unit test for this processor, which hasn't been added yet, pending review. The use case for this enhancement is the pagination links at: http://www.dreweatts.com/cms/pages/auctions?auctions.~rsp.fc[date]=p&auctions.~rsp.ss=date%2Cd.


nlevitt:
@vonrosen looks like this was a nontrivial piece of work, so I apologize, but after 4.5 years, it doesn't look like we're gonna be using it. 

Hope you're doing well!

PR #116, opened 2015-02-23
make sure log4j is configured when running unit tests, to avoid log4j er...

nlevitt:
...ror messages polluting captured output and causing tests to fail, like so: https://builds.archive.org/job/Heritrix-3/org.archive.heritrix$heritrix-commons/587/testReport/org.archive.io.arc/ARCWriterTest/testLengthTooLongCompressed/

### DIFF
--- a/commons/src/test/resources/log4j.xml
+++ b/commons/src/test/resources/log4j.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
+
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
+
+	<appender name="console" class="org.apache.log4j.ConsoleAppender">
+		<layout class="org.archive.util.OneLineSimpleLayout" />
+	</appender>
+
+	<logger name="org.apache.commons.httpclient">
+		<level value="ERROR" />
+	</logger>
+
+	<root>
+		<level value="WARN" /> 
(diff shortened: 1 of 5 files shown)


nlevitt:
This is the test failure from https://builds.archive.org/job/Heritrix-3/org.archive.heritrix%24heritrix-commons/587/testReport/org.archive.io.arc/ARCWriterTest/testLengthTooLongCompressed/
{noformat}
junit.framework.AssertionFailedError: No message log4j:WARN No appenders could be found for logger (org.archive.util.LaxHttpParser).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
WARNING Premature EOF before end-of-record: {ip-address=192.168.1.1, content-type=text/html, absolute-offset=326, subject-uri=http://www.archive.org/test/2, length=126, creation-date=20150213003141, version=null}
	at junit.framework.Assert.fail(Assert.java:47)
	at junit.framework.Assert.assertTrue(Assert.java:20)
	at org.archive.io.arc.ARCWriterTest.lengthTooLong(ARCWriterTest.java:501)
	at org.archive.io.arc.ARCWriterTest.testLengthTooLongCompressed(ARCWriterTest.java:457)
{noformat}

It's difficult to reproduce elsewhere, because of special stdout/stderr handling in the context of jenkins CI, and/or because the exact location/timing of the log4j warning message is unpredictable, I think. In any case, adding a log4j configuration should get rid of the warning messages and solve the problem.

PR #123, opened 2015-07-13
Hosts report

kris-sigur:
Addresses [HER-2084](https://webarchive.jira.com/browse/HER-2084).

I also used the opportunity to weed a bit of unused code related to hosts reporting.


nlevitt:
Thanks Kris. Only things I noticed were some mixing of tabs and spaces (heritrix convention is spaces); and properties with default values in profile-crawler-beans.cxml are normally commented out. But I figure those aren't blockers.

PR #124, opened 2015-07-30
Allow realm to be set by server for basic auth.

vonrosen:


modules/src/main/java/org/archive/modules/credential/HttpAuthenticationCredential.java
```diff
@@ -126,10 +126,18 @@ public static HttpAuthenticationCredential getByRealm(Set<Credential> rfc2617Cre
...
             for (Iterator<Credential> i = rfc2617Credentials.iterator(); i.hasNext();) {
                 HttpAuthenticationCredential c = (HttpAuthenticationCredential)i.next();
-                    if (c.getRealm().equals(realm)) {
-                        result = c;
-                        break;
-                    }
+
+                //if no realm specified, use the one returned by the server
+                if (c.getRealm() == null || c.getRealm().isEmpty()) {
+                    c.setRealm(realm);
```
nlevitt: Is it necessary to call c.setRealm()? (The http auth code isn't the easiest to follow.) It seems ugly for this method to have the side effect of modifying the credential. In the event that the crawl encounters another 401 with a different realm, then this credential won't be usable for that. If it is necessary for the realm to be set, maybe make a copy of the credential here, call setRealm(), and return that one instead.

PR #127, opened 2015-08-27
Allow User-Agent line of robots.txt to have characters before it.

vonrosen:


vonrosen:
For handling robots.txt which characters before User-Agent directive like http://thephoenix.com/robots.txt

PR #128, opened 2015-09-21
AMQPUrlReceiver changes to support RabbitMQ >= 3.3

anjackson:
As of version 3.3, [RabbitMQ no longer supports channel.flow{active=false}][1] and so the AMQPUrlReceiver module is lot able to use this method to pause message consumption.  This patch:

* Switches to a consume/cancel model so the UrlConsumer only runs when the crawl is running.
* Separates consumption from the StarterRestarter, making both more robust.
* Prevents unanticipated exceptions from killing the StarterRestarter thread.
* Enables the durability and auto-delete features of the queues to be modified via Spring.

[1]: https://www.rabbitmq.com/blog/2014/04/02/breaking-things-with-rabbitmq-3-3/

### DIFF
--- a/contrib/src/main/java/org/archive/crawler/frontier/AMQPUrlReceiver.java
+++ b/contrib/src/main/java/org/archive/crawler/frontier/AMQPUrlReceiver.java
@@ -107,9 +107,31 @@ public class AMQPUrlReceiver implements Lifecycle, ApplicationListener<CrawlStat
         return isRunning;
     }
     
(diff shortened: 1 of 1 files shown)


nlevitt:
Seems to be working for me. Thanks!

anjackson:
Thanks! I should have probably made it clear that we are running this at scale (UK domain crawl) on a almost-HEAD version of  H3 and it's working fine.

PR #132, opened 2015-10-09
try very hard to start url consumer, and therefore bind the queue to …

nlevitt:
…the routing key, so that no messages are dropped, before crawling starts (should always work unless rabbitmq is down); some other tweaks for clarity and stability


nlevitt:
@anjackson We need this change to avoid missing any urls that come back from umbra at the beginning of the crawl (in the first 10 seconds before the queue is bound). Would be great if you can review and/or test.

anjackson:
Hi, sorry for not getting back to you earlier. Had a look and tried it against our something-like-Umbra test harness and it looks good.

PR #134, opened 2015-10-30
update dist

vampireshj2013:
I think dist module's artifactId  should change from heritrix to heritrix-dist


nlevitt:
I think this would change the name of the distribution tarball to heritrix-dist...tar.gz which seems worse than the current name. Why do you think it should change?

PR #135, opened 2015-11-23
Allow KnowledgableExtractorJS to parse out youtube watch from youtube…

vonrosen:
… javascript embed

### DIFF
--- a/contrib/src/main/java/org/archive/modules/extractor/KnowledgableExtractorJS.java
+++ b/contrib/src/main/java/org/archive/modules/extractor/KnowledgableExtractorJS.java
@@ -108,7 +108,25 @@ public class KnowledgableExtractorJS extends ExtractorJS {
             }
         }
         TextUtils.recycleMatcher(m);
+        
+        //extract youtube videoid from youtube javascript embed and create link for watch page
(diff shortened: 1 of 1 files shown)


nlevitt:
Hey Hunter, looking at https://developers.google.com/youtube/iframe_api_reference?hl=en
their sample code has
```html
    <!-- 1. The <iframe> (and video player) will replace this <div> tag. -->
    <div id="player"></div>
```
```javascript
    function onYouTubeIframeAPIReady() {
        player = new YT.Player('player', {
          height: '390',
          width: '640',
          videoId: 'M7lc1UVf-VE',
          events: {
            'onReady': onPlayerReady,
            'onStateChange': onPlayerStateChange
          }
        });
      }
```

The first argument to YT.Player is an html tag id, which is arbitrary. Your regex only checks for 'ytplayer' though. Best to allow anything there.

Also, the regex should probably accept either single-quotes or double-quotes (right now it looks only for single-quotes).

Separately, it would be good to use `addRelativeToBase()`. Here's my suggested, untested code:
```java
            String newUri = "https://www.youtube.com/watch?v=" + videoId;
            try {
                addRelativeToBase(curi,
                        ext.getExtractorParameters().getMaxOutlinks(), newUri,
                        LinkContext.INFERRED_MISC, Hop.INFERRED);
            } catch (URIException e) {
                // no way this should happen
                throw new IllegalStateException(newUri, e);
            }
```

PR #136, opened 2015-11-27
Register seeds send in via AMQP

anjackson:
The original implementation allows the submitted URLs to be marked as seeds, but they are not properly registered as such and so things like SURT-prefix scoping don't pick them up.

These modifications wire the seeds in and add new seeds properly rather than just popping them in the frontier.

### DIFF
--- a/contrib/src/main/java/org/archive/crawler/frontier/AMQPUrlReceiver.java
+++ b/contrib/src/main/java/org/archive/crawler/frontier/AMQPUrlReceiver.java
@@ -37,6 +37,7 @@ import org.archive.modules.CrawlURI;
 import org.archive.modules.SchedulingConstants;
 import org.archive.modules.extractor.Hop;
 import org.archive.modules.extractor.LinkContext;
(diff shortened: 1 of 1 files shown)


nlevitt:
I don't remember being very aware of the support for isSeed and forceFetch parameters on incoming amqp messages that snuck in as part of https://github.com/internetarchive/heritrix3/pull/128/files. ;) What are you guys using these for and how are you setting them?

anjackson:
Sorry! Didn't mean to sneak anything in! I'm moving towards using the AMQP channel as a general method for setting up and feeding long-running crawls. For example, we want to run regularly crawled seeds through the web renderer before passing them to H3 (and in general want to use queues more to manage chains of processes), so it's helpful if the AMQP hook lets us submit seeds (to extend scope) and force fetching (so we can get new content links from changes in seed pages).

We can just fork the module and keep it elsewhere if it's causing problems, but I thought it might be useful for others.

nlevitt:
Cool! Glad the class is useful beyond its original intended purpose.

PR #144, opened 2016-01-16
run received urls through the candidates processor, to check scope an…

nlevitt:
…d schedule them with the frontier


nlevitt:
@anjackson does this proposed change (applying scope rules to urls received via amqp) cause any problems for you guys?

anjackson:
@nlevitt thanks for asking - I think this looks fine (I hadn't actually realised this didn't happen already!). The main feature I'm currently depending on is that I can pass a new Seed into this receiver and that these get added and passed to the SurtPrefixDecideRule so the scope grows (i.e. under seedsAsSurtPrefixes=true). AFAICT this change does not break that, so I'm happy.

nlevitt:
@anjackson 🌞great thanks

PR #146, opened 2016-01-30
More fixes for mutlipart form submission

vonrosen:
For login to https://ebiz.aacc.nche.edu/EbusPSNFYPROD/Home/Login/tabid/71/Default.aspx?returnurl=http%3a%2f%2fwww.aacc.nche.edu%2f_layouts%2fAACC.Authentication%2fLogin.aspx%3fReturnUrl%3d%252fPublications%252fCCJ%252fsubscribers%252fPages%252fdefault.aspx, the post data field of __VIEWSTATEENCRYPTED is needed or the site returns an error page and doesn't login. These changes allow blank field values to be used. Also, the submission for this page doesn't work (it returns the error page) if there are 2 submit values, Login and Validate, corresponding to the 2 buttons on the login page for Login and Email Validation. So this code change ensures that only 1 submit input is posted.

Example of blank value input needed:

--FKP3sst5b1cnB0GgQWtPpC0GkhWcKkElNs4uE1
Content-Disposition: form-data; name="__VIEWSTATEENCRYPTED"

### DIFF
--- a/modules/src/main/java/org/archive/modules/forms/HTMLForm.java
+++ b/modules/src/main/java/org/archive/modules/forms/HTMLForm.java
@@ -65,6 +65,11 @@ public class HTMLForm {
     public void addField(String type, String name, String value, boolean checked) {
         FormInput input = new FormInput();
         input.type = type;
+        
+        if (isMultipleFormSubmitInputs(type)) {
+            return;
+        }
+             
         // default input type is text per html standard
         if (input.type == null) {
             input.type = "text";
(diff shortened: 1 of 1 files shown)


nlevitt:
Merged. Did want to mention a couple things though
- comment explaining about multiple submit buttons would be good
- there could also be a <button> for submit, but not worth trying to handle that without a use case
- with this change, we also allow empty input **names** (not just values), probably harmless but a little strange

PR #147, opened 2016-02-05
If outlink is of seed that isn't in surtPrefixes, scope the outlink against the seed itself, not seeds in surtPrefixes

vonrosen:
This isn't to be merged but just for discussion purposes. The idea of this pr is to scope each outlink against the seed it was discovered from. In cases where there are multiple seeds which have different scoping restrictions, this change will prevent outlinks from restricted scope seeds from being put in scope by more leniently scoped seeds.

Example is 2 seeds in a crawl:

http://digital.lafayette.edu/collections/eastasia/
http://lafayette.edu

If the outlink http://sites.lafayette.edu/barclayp/about/ is discovered via http://digital.lafayette.edu/collections/eastasia/ it should be out of scope, even though the surt of http://lafayette.edu will put it in scope. Currently, this outlink will be in scope.

### DIFF
--- a/modules/src/main/java/org/archive/modules/deciderules/surt/SurtPrefixedDecideRule.java
+++ b/modules/src/main/java/org/archive/modules/deciderules/surt/SurtPrefixedDecideRule.java
@@ -19,6 +19,8 @@
 
 package org.archive.modules.deciderules.surt;
 
+import static org.archive.modules.CoreAttributeConstants.A_SOURCE_TAG;
+
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


anjackson:
This seems like potentially useful behaviour, but the default current behaviour of the SurtPrefixedDecideRule should not be changed. Rather it should be optional, requiring an additional property to be set, or moved to a separate class (subclass?).

ato:
Closing as this PR is now 3 years old.

I agree with Andy, I don't think most users would want this behaviour by default. I'm unclear about the motivation for it, perhaps there's a compelling reason I'm missing.

PR #148, opened 2016-02-16
Don't canonicalize source tag so that SourceSeedDecideRule will work.…

vonrosen:
… Trailing slash of canonicalized seed currently causes it not to work.


nlevitt:
This happens earlier in the method:
```
        if (!uri.matches("[a-zA-Z][\\w+\\-]+:.*")) { // Rfc2396 s3.1 scheme,
                                                     // minus '.'
            // Does not begin with scheme, so try http://
            uri = "http://" + uri;
        }
```
Source tag ought to be set to original original uri, without "http://" prepended in case it gets prepended.

PR #149, opened 2016-02-17
Update to ExtractorHTML.java for cond. comments

eleclerc:
Lot's of websites are using "Downlevel Revealed" conditional comments with a twist to have a page with valid html. this is an update to the RegEx to allow such a case. ex:

    
      HTML
    

- reference: https://css-tricks.com/downlevel-hidden-downlevel-revealed/
- reference: http://www.sitepoint.com/web-foundations/internet-explorer-conditional-comments/
- example site where this type of tag can be found: https://www.canada.ca/index.html


kris-sigur:
Thanks Eric. This resolves a comparable issue I've been dealing with in one of the more popular CMSs here in Iceland. Vote for merging.

nlevitt:
@kris-sigur I think you have the power.

PR #152, opened 2016-03-09
Switch to Java 7

anjackson:
The Heritrix codebase does not compile under Java 6 due to using [InetAddress.getLoopbackAddress()](https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getLoopbackAddress%28%29) in a test (org.archive.modules.writer.WARCWriterProcessorTest). There may be other Java 7 features in use - I've only noticed this one.

Assuming this was intentional, the Maven build needs updating to specify a source version of 1.7 or else it breaks the built in e.g. Eclipse.


nlevitt:
Cool, java 6 is old

PR #153, opened 2016-03-09
Controlling HopPath and prefetch behaviour for AMQPUrlReceiver

anjackson:
This proposal allows the prefetch behaviour to be modified, because had a problem because the AMQP message pre-fetching was too aggressive.

More importantly, this includes changes to allow the message received from AMQP to override the hop path. The default behaviour is unchanged, but if a 'hop' is set then this overrides the default value of 'I'. This also requires us to apply the high prioritisation to URLs if they are embeds or inferred (the later ensuring the original behaviour prior to this proposal is retained).

This means we can push in seeds and discovered links from seeds with the correct hop path, which makes things clearer and allows reliable prioritisation based on hop path.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+sudo: false
+
+language: java
+
+jdk:
+  - oraclejdk7
+  - oraclejdk8
+  - openjdk7
+
+matrix:
+    allow_failures:
+        - jdk: oraclejdk8
+
+before_install:  
+ - "export JAVA_OPTS=-Xmx1500m"
+ - "echo JAVA_OPTS=$JAVA_OPTS"
+ - "export MAVEN_OPTS=-Xmx1500m"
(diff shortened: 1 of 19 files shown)


anjackson:
I guess this is rather out of date now.

nlevitt:
Are you using this code? (Sorry I never got around to looking at it carefully enough to comment coherently)

anjackson:
Actually, if you don't mind, I'll re-open this for consideration since I've managed to re-sync the modifications.

This change aims to allow two things. Firstly, it allows the received URLs to control the hop path (because classifying everything as ```I``` was causing problems for us). Secondly, it allows the QOS field to be set to avoid the message consumer aggressively soaking up messages much faster than it can ACK them (which ends up starving RabbitMQ of memory for us).

anjackson:
This looks like a mess. Closing.

PR #154, opened 2016-04-02
Fixes for apparent build errors

anjackson:
Under Java 1.7 (u51), on a Mac (see full details below), I find the current build to be broken.

One set of errors arise because the unit test assumes it knows how the JVM will serialised HashMaps, and how this relates to the ordering of the HashMap. However, this ordering is not guaranteed and since some point on the 1.7 line these tests will now fail.

The other discrepancy is strange - one of the selftest crawls run during the build checks the number of novel bytes is as expected. For some reason, this value is 1 lower than it was before. In particular, the novel bytes associated with the localhost dns request appears to be 48 rather than the expected 49. Perhaps this is some kind of odd platform dependence?

~~~
$ mvn --v
Apache Maven 3.3.3 (7994120775791599e205a5524ec3e0dfe41d4a06; 2015-04-22T12:57:37+01:00)
Maven home: /usr/local/Cellar/maven/3.3.3/libexec
Java version: 1.7.0_51, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.7.0_51.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.11.3", arch: "x86_64", family: "mac"
~~~

### DIFF
--- a/engine/src/test/java/org/archive/crawler/selftest/StatisticsSelfTest.java
+++ b/engine/src/test/java/org/archive/crawler/selftest/StatisticsSelfTest.java
@@ -48,14 +48,17 @@ public class StatisticsSelfTest extends SelfTestBase {
         StatisticsTracker stats = heritrix.getEngine().getJob("selftest-job").getCrawlController().getStatisticsTracker();
         assertNotNull(stats);
         assertEquals(14, (long) stats.getCrawledBytes().get(CrawledBytesHistotable.WARC_NOVEL_URLS));
-        assertEquals(12718, (long) stats.getCrawledBytes().get(CrawledBytesHistotable.WARC_NOVEL_CONTENT_BYTES));
+        assertEquals(12717, (long) stats.getCrawledBytes()
+                .get(CrawledBytesHistotable.WARC_NOVEL_CONTENT_BYTES));
 
         assertEquals(3, (long) stats.getServerCache().getHostFor("127.0.0.1").getSubstats().get(CrawledBytesHistotable.WARC_NOVEL_URLS));
         assertEquals(2942, (long) stats.getServerCache().getHostFor("127.0.0.1").getSubstats().get(CrawledBytesHistotable.WARC_NOVEL_CONTENT_BYTES));
         assertEquals(10, (long) stats.getServerCache().getHostFor("localhost").getSubstats().get(CrawledBytesHistotable.WARC_NOVEL_URLS));
         assertEquals(9727, (long) stats.getServerCache().getHostFor("localhost").getSubstats().get(CrawledBytesHistotable.WARC_NOVEL_CONTENT_BYTES));
         assertEquals(1, (long) stats.getServerCache().getHostFor("dns:").getSubstats().get(CrawledBytesHistotable.WARC_NOVEL_URLS));
-        assertEquals(49, (long) stats.getServerCache().getHostFor("dns:").getSubstats().get(CrawledBytesHistotable.WARC_NOVEL_CONTENT_BYTES));
+        assertEquals(48,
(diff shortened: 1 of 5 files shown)


anjackson:
Thanks @ruebot - your builds hit heap space problems, but this bit looks like there are unresolved issues

~~~
Failed tests: testSomething(org.archive.crawler.selftest.StatisticsSelfTest): expected:<9775> but was:<9771>
~~~

Note that IA have their [own build server](https://builds.archive.org/job/Heritrix-3/org.archive.heritrix$heritrix-engine/618/testReport/org.archive.crawler.selftest/) which seems perfectly happy, so perhaps this a subtle platform/locale issue? 

I've attached the arc.gz from the last selftest run (from /tmp/heritrix-junit-tests/selftest/StatisticsSelfTest/jobs/selftest-job/arcs/) - can you attach yours so we can compare the contents?

[WEB-20160402195540108-00000-551~192.168.99.1~8443.arc.gz](https://github.com/internetarchive/heritrix3/files/201070/WEB-20160402195540108-00000-551.192.168.99.1.8443.arc.gz)

anjackson:
Oh that's weird. Yours doesn't have a DNS response record, and indeed no results from 'localhost' at all. Quite confused now. Maybe running out of heap space killed it midway through!?

anjackson:
Okay, I'm totally confused now. That build should not have failed like that. Maybe I should also set up a Travis build with lots of JVM combinations?

anjackson:
I eventually got Travis builds running ([e.g.](https://travis-ci.org/ukwa/heritrix3/builds/120670207)). Java 7 is fine although I kept hitting a race condition in one test and had to introduce a delay.

Java 8 hits more problems due to relying on comparing the serialised binary forms of objects. This is probably something we should try to work out how to avoid, as the serialisation format is not a fixed part of the spec, and can be expected to change from time to time and across Java implementations. However, I'm not sure how best to do this.

nlevitt:
I'm kinda letting you guys sort out these issues :)

No .travis.yml in the pull request?

anjackson:
I thought you'd rather not have one, as you already run Jenkins? Very happy to patch it in if you don't mind, as it would make it much easier for me to run Travis CI on our fork.

nlevitt:
I think it's fine to have both jenkins and travis builds. Jenkins makes it easier to download build artifacts and it populates our maven repo. We can use travis for testing with different versions of java and stuff like that.

anjackson:
Thanks to @ruebot for finishing off the Travis config.

kris-sigur:
Been using the travis build against one of my [forks](https://github.com/Landsbokasafn/heritrix3/tree/LBS-2016-02). The OpenJDK build seems to occasionally fail (different unit tests) for no (seemingly) good reason? The OracleJDK, however, has only failed when it should have.

nlevitt:
Sorry this slipped off my radar. @ruebot I cherry-picked the .travis.yml commit so we can have a baseline before merging these fixes. https://travis-ci.org/internetarchive/heritrix3

nlevitt:
I didn't have to do anything in github settings. (Maybe there was some account level setting that was already enabled though.) I just enabled it in travis.

nlevitt:
After running into more issues with those byte count checks in StatisticsSelfTest.java I finally took the time to examine the issue. Turns out the size of the dns record depends on the local environment (dns server, I guess). Specifically the TTL varies. For example
```
20160606204018
localhost.              0       IN      A       127.0.0.1
```

vs.

```
20160606200119  
localhost.              604800  IN      A       127.0.0.1
```

On #164 I extended @anjackson's branch to exclude dns records from the size checks.

PR #164, opened 2016-06-06
Fixes for apparent build errors (extends #154)

nlevitt:


nlevitt:
See #154

kris-sigur:
There is something about the OpenJDK 7 build that causes it to sometimes fail and sometimes succeed. 

I have, however, yet to see the OracleJDK 7 fail without good reason.

nlevitt:
🐵 Tests passed on all jdks.

Anybody worried about performance hit of switching from HashMap/HashSet/ConcurrentHashMap to TreeMap/TreeSet/ConcurrentSkipListMap? I'm thinking none of the objects we're changing are big enough for it to matter.

PR #169, opened 2016-08-20
Add annotation for write reject decisions.

BitBaron:


nlevitt:
Here's an excerpt from Processor.java. This is where the shouldProcessRule is enforced. I think WritePoolProcessor should override innerRejectProcess and set the annotation there.
<pre>
    public ProcessResult process(CrawlURI uri) 
    throws InterruptedException {
        if (!getEnabled()) {
            return ProcessResult.PROCEED;
        }
        
        <b>if (getShouldProcessRule().decisionFor(uri) == DecideResult.REJECT) {
            innerRejectProcess(uri);
            return ProcessResult.PROCEED;
        }</b>
        
        if (shouldProcess(uri)) {
            uriCount.incrementAndGet();
            return innerProcessResult(uri);
        } else {
            return ProcessResult.PROCEED;
        }
    }
</pre>

Good use of `ANNOTATION_UNWRITTEN`. The other part of the annotation could be problematic though. I think `getShouldProcessRule().getClass()` will end up producing a string like `"class org.archive.modules.deciderules.DecideRuleSequence"`, with a space in it, which is a problem. Also including the name of the class there doesn't really tell you very much. I think better would be `"unwritten:rejected(shouldProcessRule)"`.
```
+            curi.getAnnotations().add(ANNOTATION_UNWRITTEN + ":rejected("
+                    + getShouldProcessRule().getClass() + ")");
```

PR #171, opened 2016-10-11
AMQP fine control

anjackson:
This change aims to allow two things. Firstly, it allows the received URLs to control the hop path (because classifying everything as I was causing problems for us). Secondly, it allows the QOS field to be set to avoid the message consumer aggressively soaking up messages much faster than it can ACK them (which ends up starving RabbitMQ of memory for us).

I also updated Kryo 1 to latest bug-fix version, but this can probably be taken out if you'd rather.

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -161,9 +161,9 @@
 		</dependency>
 
 		<dependency>
-			<groupId>com.googlecode</groupId>
+			<groupId>com.esotericsoftware</groupId>
 			<artifactId>kryo</artifactId>
-			<version>1.04</version>
+			<version>1.01</version>
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -163,7 +163,7 @@
 		<dependency>
 			<groupId>com.esotericsoftware</groupId>
 			<artifactId>kryo</artifactId>
-			<version>1.01</version>
+			<version>1.04</version>
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -161,7 +161,7 @@
 		</dependency>
 
 		<dependency>
-			<groupId>com.esotericsoftware</groupId>
+			<groupId>com.googlecode</groupId>
 			<artifactId>kryo</artifactId>
 			<version>1.04</version>
 			<scope>compile</scope>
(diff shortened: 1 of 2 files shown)


commons/pom.xml
```diff
@@ -161,9 +161,9 @@
 		</dependency>
 
 		<dependency>
-			<groupId>com.esotericsoftware</groupId>
+			<groupId>com.googlecode</groupId>
```
nlevitt: This looks like a good change, but was it meant to be part of this pull request?

contrib/src/main/java/org/archive/crawler/frontier/AMQPUrlReceiver.java
```diff
@@ -152,6 +151,28 @@ public void setForceFetch(boolean forceFetch) {
         this.forceFetch = forceFetch;
     }
 
+    /**
+     * The maximum prefetch count to use, meaning the maximum number of messages
+     * to be consumed without being acknowledged. Use 'null' to specify there
+     * should be no upper limit (the default).
+     */
+    private Integer prefetchCount = null;
```
nlevitt: Does this really need to be configurable or can we use a sensible default?

anjackson: It doesn't need to be configurable, I just assumed that would be preferred. A default of e.g. 1000 or something would prevent the problems we've seen.

contrib/src/main/java/org/archive/crawler/frontier/AMQPUrlReceiver.java
```diff
@@ -409,14 +434,27 @@ protected CrawlURI makeCrawlUri(JSONObject jo) throws URIException,
...
              */
-            curi.setSchedulingDirective(SchedulingConstants.HIGH);
-            curi.setPrecedence(1);
+            if (Hop.INFERRED.getHopString().equals(curi.getLastHop())
+                    || Hop.EMBED.getHopString().equals(curi.getLastHop())) {
+                curi.setSchedulingDirective(SchedulingConstants.HIGH);
+                curi.setPrecedence(1);
+            } else {
+                /*
+                 * By default, redirects get set to MEDIUM, so to ensure the
```
anjackson: This looks like noise from an earlier iteration, I'll drop it.

contrib/src/main/java/org/archive/crawler/frontier/AMQPUrlReceiver.java
nlevitt: 1000 sounds good to me!

anjackson:
‘E’s and ‘L’s and ‘R’s will pick up the default behaviour from the FrontierPreparer calling innerProcess > prepare > curi.setSchedulingDirective(getSchedulingDirective(curi)) so there's no need to handle those usual cases here.

anjackson:
Not sure why that test failed - seems to be unrelated:

```
Failed tests:   testConcurrentLoad(org.archive.modules.fetcher.CookieStoreTest)
```

PR #179, opened 2017-05-04
Add parsing for srcset attributes

BitBaron:
Add parsing for srcset attributes
Include Unit Tests for srcset cases


nlevitt:
Thanks Neil!

PR #192, opened 2017-10-31
Enforce robots.txt character limit per char not per line

ato:
Since we were only enforcing the size limit after reading each
line a single very long line could cause us to to run out of
memory.

This change enforces the character limit on each read character
not just on line boundaries.

We also correct the count of processed characters in the warning
message, which was not counting newline and linefeed characters.

### DIFF
--- a/modules/src/main/java/org/archive/modules/net/CrawlServer.java
+++ b/modules/src/main/java/org/archive/modules/net/CrawlServer.java
@@ -174,10 +174,9 @@ public class CrawlServer implements Serializable, FetchStats.HasFetchStats, Iden
 
         InputStream contentBodyStream = null;
         try {
-            BufferedReader reader;
             contentBodyStream = curi.getRecorder().getContentReplayInputStream();
 
-            reader = new BufferedReader(new InputStreamReader(contentBodyStream));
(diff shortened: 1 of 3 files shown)


gojomo:
Might be simpler to reuse something like org.apache.commons.io.input.BoundedReader, with a limit of MAX_SIZE+1, log overflow when that one-char excess noticed.

ato:
That was my first thought too. However, how would we know when we reach the limit (vs end of file)? Counting the length of the returned strings like the original code does is inaccurate as it can't distinguish between 1 char LF, 2 char CRLF endings 0 char and end of file. Test reading a byte from the underlying reader doesn't work if we're still wrapping with BufferedReader as it reads ahead filling the buffer. Is there a non-buffered readLine somewhere else we could use? 

If we don't accurately know when we reach the size limit vs end of file we risk processing a truncated line. Maybe that's not as big a deal as I think it is, I'm imaging 'Disallow /foo' being truncated to 'Disallow /'. Which seems worse to me than ignoring the problem remaining lines. But could be I'm overthinking things.

ato:
Hahah yep, of course I'm over thinking it. Noah on IIPC Slack: read 500k and split the string.

Issue #208, opened 2018-07-04
HTML extractor does not handle the base href correctly when it's relative


anjackson:
From archive-crawler:

>On Tue, Jul 3, 2018 at 3:51 AM david.reiner82@yahoo.com [archive-crawler] <archive-crawler@yahoogroups.com> wrote:
>
> At the moment Heritrix (3.3.0 Snapshop) seems to ignore base href tags. 
>
> Example:
https://www.schmid-gartenpflanzen.de/forum/index.php/mv/msg/7627/216142/0/
> <base href="/forum/"/>
> <a href="/forum/index.php/fa/89652/0/" title="IMG_7651kl.jpg">IMG_7651kl.jpg</a>
> 
> Web browsers interpret the link this way:
> https://www.schmid-gartenpflanzen.de/forum/index.php/fa/89652/0/
> 
> However, Heritrix calls this url:
> https://www.schmid-gartenpflanzen.de/forum/index.php/mv/msg/7627/216142/0/index.php/fa/89652/0/
> 
> This behaviour leads to thousands of false urls being crawled. From my point of view, the Heritrix behavior is a bug, isn't it?

If I construct a test case with a relative base href, I get:

```
INFO: https://www.schmid-gartenpflanzen.de/forum/index.php/mv/msg/7627/216142/0/
org.apache.commons.httpclient.URIException: Relative URI but no base: /forum/
	at org.archive.url.UsableURIFactory.fixup(UsableURIFactory.java:416)
	at org.archive.url.UsableURIFactory.create(UsableURIFactory.java:275)
	at org.archive.url.UsableURIFactory.create(UsableURIFactory.java:265)
	at org.archive.net.UURIFactory.getInstance(UURIFactory.java:44)
	at org.archive.modules.extractor.ExtractorHTML.processGeneralTag(ExtractorHTML.java:426)
	at org.archive.modules.extractor.ExtractorHTML.extract(ExtractorHTML.java:850)
	at org.archive.modules.extractor.ExtractorHTMLTest.testRelativeBaseHrefRelativeLinks(ExtractorHTMLTest.java:284)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

So I think this is a bug.

PR #209, opened 2018-07-04
Test case and fix for internetarchive/heritrix3#208.

anjackson:
I believe this ensures Heritrix3 deals with relative base hrefs.

### DIFF
--- a/modules/src/test/java/org/archive/modules/extractor/ExtractorHTMLTest.java
+++ b/modules/src/test/java/org/archive/modules/extractor/ExtractorHTMLTest.java
@@ -270,6 +270,54 @@ public class ExtractorHTMLTest extends StringExtractorTestBase {
     }
     
     /**
+     * Test that relative base href's are resolved correctly:
+     * 
+     * See 
+     * 
+     * @throws URIException
+     */
+    public void testRelativeBaseHrefRelativeLinks() throws URIException {
+        CrawlURI curi = new CrawlURI(UURIFactory
+                .getInstance("https://www.schmid-gartenpflanzen.de/forum/index.php/mv/msg/7627/216142/0/"));
+        CharSequence cs = "<base href=\"/forum/\"/>\n" + 
+                "<img src=\"index.php/fa/89652/0/\" border=\"0\" alt=\"index.php/fa/89652/0/\" />";
+        getExtractor().extract(curi, cs);
+
+        assertTrue(CollectionUtils.exists(curi.getOutLinks(), new Predicate() {
+            public boolean evaluate(Object object) {
+                return ((CrawlURI) object)
+                        .getURI()
+                        .indexOf(
+                                ".de/forum/index.php/fa/89652/0/") >= 0;
+            }
+        }));
+    }
+
+    /**
+     * Test that absolute base href's are resolved correctly:
+     * 
+     * @throws URIException
+     */
+    public void testAbsoluteBaseHrefRelativeLinks() throws URIException {
+
+        CrawlURI curi = new CrawlURI(UURIFactory
+                .getInstance("https://www.schmid-gartenpflanzen.de/forum/index.php/mv/msg/7627/216142/0/"));
+        CharSequence cs = "<base href=\"https://www.schmid-gartenpflanzen.de/forum/\"/>\n" + 
+                "<img src=\"index.php/fa/89652/0/\" border=\"0\" alt=\"index.php/fa/89652/0/\" />";
+        getExtractor().extract(curi, cs);
+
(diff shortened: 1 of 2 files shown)


anjackson:
Hm, well, this now breaks other tests. Elsewhere, [BAD BASE](https://github.com/internetarchive/heritrix3/search?q=%22bad+base%22&unscoped_q=%22bad+base%22) tests are in play that assume `base` `href` should be absolute.

Not sure what to do about this.

ato:
Hm. So I think the problem is that [HTML4 specified](https://www.w3.org/TR/html4/struct/links.html#edef-BASE) that base URLs must be absolute. Presumably browsers at the time ignored absolute URLs hence the BADBASE tests. However [HTML5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) allows them to be relative but [only the first is obeyed](https://html.spec.whatwg.org/multipage/semantics.html#the-base-element) and any subsequent base element is ignored.

So I guess the options for Heritrix's behaviour are:
1. Follow what contemporary browsers do: allow relative URLs and ignore base tags other than the first.
2. Generate candidates for every possible case.
3. Change behaviour depending on the doctype.

Personally I would go with option 1 since it's simple and matches what browsers do. I would consider option 2 speculative link extraction.

anjackson:
I'm happy to implement option 1. Option 2 gets quite messy as it would mean storing all potential base URIs and generating all potential URL combinations (i.e. for _L_ links and _B_ base URIs we'd need to emit _L x B_ links!) Option 3 is doable too I guess, with a fallback to option 1 when the DOCTYPE is not known.

I'm not really sure what to do with the tests, as they appear to be testing what happens when you get bad URIs, and happen to use the HTML4 base URI convention to make it happen. Having taken that away, I am struggling to make a test case that actually does create a bad URI!

ato:
> I'm not really sure what to do with the tests, as they appear to be testing what happens when you get bad URIs

If we're going with option 1 I think we should replace those bad url tests entirely. They're just plain wrong for modern html. We should instead test relative base hrefs and a page with several base tags.

anjackson:
The problem is that the failing test is called `BadURIsStopPageParsingSelfTest` i.e. it's about hitting bad URIs (that it uses base hrefs to do it seems to be incidental). Having modified the base href handling, I don't know how to make the `BadURIs` test go off. I guess we could use very, very long URIs? If I add 'bad' characters it just escapes them.

ato:
Looks like this is the original issue: https://webarchive.jira.com/browse/HER-25


> I don't know how to make the BadURIs test go off.

I'm confused, why would we want them to? Isn't the point of the original issue that Heritrix should NOT choke on bad URIs? (ie the name of the test is describing the problem addressed not the desired behaviour)

ato:
Those tests look all wrong to me. I'm pretty sure modern browsers would apply the base urls in all cases (one.html two.html html). If there's extra slashes or illegal characters ir whatever modern browsers just normalise and escape them. The schemaless url is treated as a relative URL. So I still think the tests are wrong and should be removed. (Or inverted so they enforce that goodone.html etc are actually not found)

anjackson:
Yes, and the test appears to be designed to check that the page parsing does not stop if the extractor hits a bad URI. If I can't create a URI error, then I can't make the test test anything.

anjackson:
Okay, so from a Real Life Log File^(TM) I find three types of error:

- URI length > 2083
- URI too short to be a meaningful URI
- Contains non-LDH characters due to a `%20` in the host name.

I think the latter is probably easiest to use.

anjackson:
Right, so, finally got the tests working. One of the other tests assumed BaseURI would be set every time a `base` element was read, rather than just the first, so I had to modify that too.

modules/src/test/java/org/archive/modules/extractor/ExtractorHTMLTest.java
```diff
@@ -329,7 +402,6 @@ public void testScriptTagWritingScriptType() throws URIException {
     public void testOutLinksWithBaseHref() throws URIException {
         CrawlURI puri = new CrawlURI(UURIFactory
                 .getInstance("http://www.example.com/abc/index.html"));
-        puri.setBaseURI(puri.getUURI());
```
anjackson: This was the change needed because BaseURI can now only be set once in ExtractorHTML

kris-sigur:
@anjackson It looks fine, but unfortunately you caught me minutes before I'm out the door for a holiday so I can't properly vet it until the end of the month.

anjackson:
Hey @nlevitt @kris-sigur or anyone able to take a quick look at this?

PR #213, opened 2018-09-14
Ensure frontier work queues are updated safely across threads.

### LINKED ISSUE #212: BdbFrontier thread safety
We're attempting to use Heritrix3 with an external module that populates the BdbFrontier via Kafka, and we're hitting problems interacting with the frontier safely. There's some more details in ukwa/ukwa-heritrix#16, but to summarise, ToeThreads are dying because [`keepItem`](https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/engine/src/main/java/org/archive/crawler/frontier/WorkQueue.java#L99) is `null` when it should not be.

I believe this is because `peekItem` is marked as `transient`. Occasionally, between setting `peekItem` ([this statement](https://github.com/internetarchive/heritrix3/blob/3892c17b56c409c10e46eb0951ca18296d3d31f2/engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java#L926)) and using it ([this one](https://github.com/internetarchive/heritrix3/blob/3892c17b56c409c10e46eb0951ca18296d3d31f2/engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java#L947)), the `WorkQueue` gets updated by a separate thread in a way that forces it to get written out to disk and then read back in again. As `peekItem` is `transient`, flushing it out to the disk and back drops the value and we're left with a `null`.

[NetArchive Suite have also seen this issue](https://sbforge.org/jira/browse/NAS-2560) when using a RabbitMQ-based URL receiver, and [patched it by ignoring the `null`](https://github.com/netarchivesuite/heritrix3/commit/cb087b154c9fa80ef902095dea37d40886bd080e).

The simplest way to avoid this would be to remove the `transient` modified from `peekItem` but that makes me worry because someone deliberately chose to make it `transient` and I don't understand why.

Secondly, I don't understand why we are seeing this, when IA also use similar methods and are (presumably?) not seeing this. Moreover, this model appears not to be fundamentally different to the traditional `ActionDirectory`, so I don't understand why this wasn't seen a long time ago.

Finally, this issue also made it clear that I don't actually understand how best to interact with the BdbFrontier in a thread-safe manner.  If I am right in assuming that every modification to a `WorkQueue` needs to be followed by a `.makeDirty()` that serialised the queue out to disk and reads it back in again, then surely every modification needs to edit-then-write within a `synchronized(WorkQueue)` block? But it's pretty easy to find examples where this appears to be deliberately not the case:

https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java#L390-L410

I'd appreciate any information anyone has on how best to inject URLs into Heritrix3, and on whether or not I've understood how the `BdbFrontier` works.

anjackson:
In an attempt to resolve #212, wrapping all WorkQueue updates in `synchronized` blocks. Has been tested at moderately large scale running the UKWA frequent crawls.

### DIFF
--- a/engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
+++ b/engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
@@ -602,17 +602,17 @@ implements Closeable,
                         }
                     }
                     readyQ = getQueueFor(key);
-                synchronized (readyQ) {
                     if (readyQ == null) {
                         // readyQ key wasn't in all queues: unexpected
                         logger.severe("Key " + key
                                 + " in readyClassQueues but not allQueues");
                         break findaqueue;
                     }
+                synchronized (readyQ) {
                     if (readyQ.getCount() == 0) {
                         // readyQ is empty and ready: it's exhausted
-                            readyQ.noteExhausted(); 
-                            readyQ.makeDirty();
+                        readyQ.noteExhausted();
+                        readyQ.makeDirty();
                         readyQ = null;
                         continue;
                     }
--- a/engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
+++ b/engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
@@ -404,10 +404,10 @@ implements Closeable,
                     deactivateQueue(wq);
                 }
             }
+            // Update recovery log.
+            doJournalAdded(curi);
+            wq.makeDirty();
         }
-        // Update recovery log.
-        doJournalAdded(curi);
-        wq.makeDirty();
         largestQueues.update(wq.getClassKey(), wq.getCount());
     }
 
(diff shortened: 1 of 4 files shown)


anjackson:
I've had to muck around a bit to make Travis CI happy, but I think this works fine.

I believe this also resolves:

- https://webarchive.jira.com/browse/HER-507
- https://sbforge.org/jira/browse/NAS-2560

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
```diff
@@ -600,54 +602,57 @@ protected CrawlURI findEligibleURI() {
                         }
                     }
                     readyQ = getQueueFor(key);
-                    if(readyQ==null) {
-                         // readyQ key wasn't in all queues: unexpected
-                        logger.severe("Key "+ key +
-                            " in readyClassQueues but not allQueues");
+                synchronized (readyQ) {
+                    if (readyQ == null) {
```
ato: This branch is unreachable as synchronised will NullPointerException. Should do the null check before synchronised.

anjackson: I think I've resolved this now.

anjackson:
If I don't hear any objections I'm going to go ahead and merge this in day or two. /cc @nlevitt @kris-sigur

anjackson:
Sorry, I mean, I'll merge it on the basis that @ato has reviewed it and we (UKWA) have run it at scale and it performs well.

PR #220, opened 2018-11-15
implement PredicatedDecideRule.onlyDecision()

nlevitt:
DecideRuleSequence already has the optimization that I was looking for,
namely, don't bother evaluating a DecideRule if we know it won't change
the current result. For some reason PredicatedDecideRule, which is a
parent class to most of the decide rules in heritrix, didn't implement
onlyDecision(). Certain crawl configurations could see significant
performance improvement with this change.

### DIFF
--- a/modules/src/main/java/org/archive/modules/deciderules/PredicatedDecideRule.java
+++ b/modules/src/main/java/org/archive/modules/deciderules/PredicatedDecideRule.java
@@ -28,6 +28,7 @@ import org.archive.modules.CrawlURI;
  * @author gojomo
  */
 public abstract class PredicatedDecideRule extends DecideRule {
+    private static final long serialVersionUID = 1L;
 
     {
         setDecision(DecideResult.ACCEPT);
@@ -48,6 +49,11 @@ public abstract class PredicatedDecideRule extends DecideRule {
             return getDecision();
         }
         return DecideResult.NONE;
+    }
+
+    @Override
+    public DecideResult onlyDecision(CrawlURI uri) {
+        return getDecision();
     }
 
     protected abstract boolean evaluate(CrawlURI object);


anjackson:
This is only fine if you assume the decide rules do not have any side-effects. Unfortunately, that's not always the case! As such, while this is a much more sensible default setting, this is potentially quite a significant behavioural change, so should be noted clearly in the change log.

(we have a couple of examples, but we have already added a `lookupEveryUri` parameter to control this behaviour)

nlevitt:
Oh sorry, didn't consider that!

What's the `lookupEveryUri` parameter?

One way to solve for a given PredicatedDecideRule class would be to override `onlyDecision()` to return null;

anjackson:
Ah I thought we’d got this from core Heritrix but it looks like it’s an UKWA thing: https://github.com/ukwa/ukwa-heritrix/blob/56fc4ecd03882c54f180e44409551f8ed82cf96b/src/main/java/uk/bl/wap/modules/deciderules/ExternalGeoLocationDecideRule.java#L63

We may be the only ones who use such foul tactics?

ato:
ExternalGeoLocationDecideRule (minus Andy's lookupEveryUri addition) is indeed a stock Heritrix rule and does also have the side-effect of stashing the country code in CrawlHost. However nothing else in stock Heritrix reads it and it appears originally intended to just be a cache for the rule itself. 

Maybe we should more strongly encourage the upstreaming of changes like the UKWA's and the NetArchiv suite's? I don't think it's fair for the upstream Heritrix project to have to worry about breaking out of tree code. It has a paralysing effect where those making changes have to worry if someone might possibly be relying on each and every detail. I realise Andy isn't saying 'don't make the change' but rather 'document it' but even so.

That would also give us the opportunity to explore alternative solutions - either upfront when the upstreaming happens or down the track when someone wants to make a change like this PR. For example in this case I'd wonder if rather than relying on the side effect of the decide rules it might be possible to factor out the side-effect code into a common method which both the decide rule and whatever else uses the country code can call. That'd ensure its available in both places without forcing the decide rule to always be run.

Issue #229, opened 2019-02-03
Make FetchHistoryProcessor 304 handler more robust


anjackson:
The `FetchHistoryProcessor`  has special logic to handle HTTP 304 responses...

https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/modules/src/main/java/org/archive/modules/recrawl/FetchHistoryProcessor.java#L109-L126

...but we're having problems because a server is returning this even though we never sent a `If-Modified-Since` or `If-None-Match` in the request (pretty sure the crawler never does this).

I'm not sure this is an appropriate use of de-duplication logic: de-duping a 304 because the server tells us it's the same as an earlier response is really not the same as de-duping a 200 that we've downloaded before and de-duping after verifying the hashes match.

So, the whole section should be removed, IMO.  Failing that, we'll need to re-write it so if does not assume the history is present.

(see https://github.com/ukwa/ukwa-heritrix/issues/27 for context)

anjackson:
Ah, it looks like some of the IA HBase code uses A_ETAG_HEADER: https://github.com/internetarchive/heritrix3/search?q=A_ETAG_HEADER&unscoped_q=A_ETAG_HEADER

Well, as much as I dislike trusting the original servers, probably easiest to just check the history exists to about the NPE.

anjackson:
Resolved by #230

PR #230, opened 2019-02-04
Add checks to guard against server sending 304 in error

anjackson:
Following #229, this small change avoids problems when a server sends a 304 response in error. De-duplication will not be attempted unless the digest is there to de-duplicate with.


anjackson:
@ato Thanks!

Issue #221, opened 2018-11-21
ToeThread death when using HighestUriPrecedenceProvider


anjackson:
We're using `HighestUriPrecedenceProvider` and have noticed this occasional (very rare) failure leading to ToeThreads dying...

```
SEVERE: org.archive.crawler.framework.ToeThread run Fatal exception in ToeThread #53:  [Sat Nov 03 14:36:05 GMT 2018]
java.util.NoSuchElementException
        at java.util.TreeMap.key(TreeMap.java:1327)
        at java.util.TreeMap.firstKey(TreeMap.java:290)
        at org.archive.crawler.frontier.precedence.HighestUriQueuePrecedencePolicy$HighestUriPrecedenceProvider.getPrecedence(HighestUriQueuePrecedencePolicy.java:89)
        at org.archive.crawler.frontier.WorkQueue.getPrecedence(WorkQueue.java:627)
        at org.archive.crawler.frontier.WorkQueueFrontier.activateInactiveQueue(WorkQueueFrontier.java:781)
        at org.archive.crawler.frontier.WorkQueueFrontier.findEligibleURI(WorkQueueFrontier.java:597)
        at org.archive.crawler.frontier.AbstractFrontier.next(AbstractFrontier.java:450)
        at org.archive.crawler.framework.ToeThread.run(ToeThread.java:134)
```

Presumably this is likely some subtle race-condition, i.e. the TreeMap being modified while it's being read.

See https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/engine/src/main/java/org/archive/crawler/frontier/precedence/HighestUriQueuePrecedencePolicy.java#L89

anjackson:
Googling indicates this can happen [because TreeSet will merge elements when multiple entries have the same value](https://stackoverflow.com/a/26305231), so that might be what's happening.

anjackson:
The [API spec says](https://docs.oracle.com/javase/7/docs/api/java/util/TreeMap.html#firstKey()) that `firstKey` will throw `NoSuchElementException` if it's empty, so presumably this is a subtle issue where enqueuedCounts is being emptied between the `size() > 0` check and the `.firstKey()` access.

The belt-and-braces approach would be to wrap all accessed to `enqueuedCounts` in this class in a `synchronised(enqueuedCounts)` block.

ato:
It looks like it's slipping through the WorkQueue locks in several places. This is the one in your backtrace:

```java
[...]

PR #231, opened 2019-02-04
Add synchronized statements for internetarchive/heritrix3#221.

anjackson:
This wraps all accesses to `enqueuedCounts` in this class `synchronized`, to avoid the race-condition observed in internetarchive/heritrix3#221


anjackson:
@ato Thanks!

PR #241, opened 2019-03-14
Fix for srcset parsing base64 encoded images as urls.

adam-miller:
srcset attributes contain a comma separated list of urls, however base64 encoded images also contain a comma. This causes the srcset url parser to split the encoded image and interpret it as 2 urls instead of ignoring it.

### DIFF
--- a/modules/src/main/java/org/archive/modules/extractor/ExtractorHTML.java
+++ b/modules/src/main/java/org/archive/modules/extractor/ExtractorHTML.java
@@ -685,6 +685,11 @@ public class ExtractorHTML extends ContentExtractor implements InitializingBean 
             String[] links = value.toString().split(",");
             for (int i=0; i < links.length; i++){
                 String link = links[i].trim().split(" +")[0];
+		//base64 encoded images contain a comma, so the next element is a continuation of this encoded image
+		if(link.startsWith("data:image")){
+		    i++;
+		    continue;
+		}
                 logger.finer("Found " + link + " adding to outlinks.");
                 addLinkFromString(curi, link, context, hop);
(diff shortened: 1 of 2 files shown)


ato:
Looking at the [parsing algorithm](https://html.spec.whatwg.org/multipage/images.html#parsing-a-srcset-attribute) in the spec I think we might need something more complicated to get this right. Looks like commas are allowed in the middle of any URL (not just data:) and also allowed between parentheses in the descriptors.

Trying to translate the WHATWG algorithm into a regexs I got this:

```java
        Pattern SRCSET_URL = Pattern.compile("[\\s,]*(\\S+)");
        Pattern SRCSET_DESCRIPTORS = Pattern.compile("\\s*(?:[^,(]+|\\([^)]*\\))*");

        Matcher m = SRCSET_URL.matcher(value);
        while (m.lookingAt()) {
            String link = m.group(1);
            m.region(m.end(), m.regionEnd());
            
            if (link.endsWith(",")) {
                // URLs that end in comma have no descriptors
                link = link.replaceFirst(",+$", "");
            } else {
                // skip over descriptors
                m.usePattern(SRCSET_DESCRIPTORS);
                if (!m.lookingAt()) {
                    throw new IllegalStateException("bug: non-match should be impossible");
                }
                m.region(m.end(), m.regionEnd());
                m.usePattern(SRCSET_URL);
            }

            logger.finer("Found " + link + " adding to outlinks.");
            addLinkFromString(curi, link, context, hop);
            numberOfLinksExtracted.incrementAndGet();
        }
```

nlevitt:
Wow @ato. We were just discussing this and were about to give up on actually doing it fully correctly. Thanks.

PR #243, opened 2019-03-16
Handle commas more compliantly when parsing srcset

ato:
Commas are allowed if they're in the middle of the URL. Consequently:

    srcset="a,b,,c,"   => ["a,b,,c"]
    srcset="a, b,, c," => ["a", "b", "c"]

They occur particularly commonly in data: URLs before the base64 value.

Commas are also allowed in descriptors if they are enclosed by parens:

    srcset="a (b,c),d" => ["a", "d"]

Spec: https://html.spec.whatwg.org/multipage/images.html#parsing-a-srcset-attribute

### DIFF
--- a/modules/src/main/java/org/archive/modules/extractor/ExtractorHTML.java
+++ b/modules/src/main/java/org/archive/modules/extractor/ExtractorHTML.java
@@ -682,7 +682,7 @@ public class ExtractorHTML extends ContentExtractor implements InitializingBean 
 
             logger.fine("Found srcset listing: " + value.toString());
 
-            Matcher matcher = TextUtils.getMatcher("[\\s,]*(\\S*[^,\\s])(?:\\s(?:[^,(]+|\\([^)]*\\))*)?", value);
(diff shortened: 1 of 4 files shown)


ato:
@nlevitt, @adam-miller, this is a simplified version of my comment in #241. This should be pretty close to spec. I tried the test string against https://github.com/albell/parse-srcset and got the same result.

herrernst:
Should this fix the images on http://web.archive.org/web/*/https://orf.at/ ? Unfortunately, they are all broken at the moment, because a `,` is used in the individual image URLs and they are rewritten incorrectly.

ato:
> Should this fix the images on http://web.archive.org/web/*/https://orf.at/ ?

No, Heritrix is not involved in replay URL rewriting. As far as I'm aware the new version of the replay software used on web.archive.org is not open source.

herrernst:
Thanks. Do you know where I can report this issue then? Or could you forward it please? Thanks.

ato:
I believe emailing info@archive.org is the appropriate way to contact the Internet Archive. (I don't work for them, I'm just one of the community members that helps maintain Heritrix.)

Issue #245, opened 2019-03-20
Upgrade HTTP Client to 4.5.x


anjackson:
We're seeing some minor issues (ukwa/ukwa-heritrix#31) that appear to be bugs in the HTTP Client, so I'd like to see about upgrading.  I've attempted to build with 4.5.7 and it compiles fine, but the cookie-related tests fail with errors like `iterator() not implemented`, and indeed the iterator is not:

https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/modules/src/main/java/org/archive/modules/fetcher/BdbCookieStore.java#L81

As per the above comment:

https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/modules/src/main/java/org/archive/modules/fetcher/BdbCookieStore.java#L66-L73

But the HTTP Client code now expects an iterator, so we need to figure out a way of providing an iterator interface but ensuring the underlying iterator gets closed.

anjackson:
Not sure how best to handle this. Could we wrap the `StoredCollection` in an iterator class that adds a `finalize` method to close the cursor on GC? And/or monitor when the iterator runs out and close it before handing back control?

anjackson:
Also I think the Danish web archive team were looking at upgrading BDB, and maybe later versions offer different options?

ato:
Note that finalizers were deprecated in jdk 9 and are removed entirely from the language in jdk 12.

ato:
Given the only method the existing code implements apparantly is toArray() it seems like `Arrays.asList(toArray()).iterator()` should do

anjackson:
Well that does allow the tests to run, and most of them work. The one that doesn't may be a faulty test.
 
https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/modules/src/test/java/org/archive/modules/fetcher/CookieFetchHTTPIntegrationTest.java#L324-L330

anjackson:
See https://github.com/ukwa/heritrix3/tree/upgrade-httpclient for experimental branch.

ato:
Per RFC6265:
> Unless the cookie's attributes indicate otherwise, the cookie is returned only to the origin server (and not, for example, to any subdomains),

So that subdomain test is indeed now  incorrect behaviour.

anjackson:
Having run this for a while, this problem turned up:

```
[...]

PR #246, opened 2019-03-21
Upgrade httpclient to 4.5.7 and handle cookies more compliantly

anjackson:
Following #245 this updated to version 4.5.7 of HTTP Client. Currently, this has only been validated using the built-in tests. One test was outdated and was removed.

### DIFF
--- a/modules/src/main/java/org/archive/modules/fetcher/BdbCookieStore.java
+++ b/modules/src/main/java/org/archive/modules/fetcher/BdbCookieStore.java
@@ -20,6 +20,7 @@ package org.archive.modules.fetcher;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
 import java.util.Iterator;
@@ -78,7 +79,7 @@ public class BdbCookieStore extends AbstractCookieStore implements
         @Override public int size() { return wrapped.size(); }
         @Override public boolean isEmpty() { throw new RuntimeException("not implemented"); }
         @Override public boolean contains(Object o) { throw new RuntimeException("not implemented"); }
-        @Override public Iterator<T> iterator() { throw new RuntimeException("not implemented"); }
+        @Override public Iterator<T> iterator() { return (Iterator<T>) Arrays.asList(wrapped.toArray()).iterator(); }
         @Override public Object[] toArray() { return wrapped.toArray(); }
         @SuppressWarnings("hiding") @Override public <T> T[] toArray(T[] a) { return wrapped.toArray(a); }
         @Override public boolean add(T e) { throw new RuntimeException("immutable list"); }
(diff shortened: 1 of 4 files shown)


anjackson:
I'd ideally like to hear from IA (ping @nlevitt) in case anyone has objections to this going in at the moment?  I think it's a pretty straightforward upgrade, but I'd like to ensure they are at least aware of it, given how core this dependency is to H3 actually doing anything at all!

nlevitt:
I've been following the discussion but I haven't had a chance to do my homework. Since you asked, I'll just pose the questions I have in mind:

1. Have we looked at performance implications of doing `toArray()` frequently?
2. Is the subdomain thing the only implication of switching from `CookieSpecs.BROWSER_COMPATIBILITY` to `CookieSpecs.DEFAULT`? I guess it's the only one that came up in testing?
3. I think RFC6265 existed when I wrote that comment ("XXX I think browsers differ on this behavior"), but browser behavior is the real spec heritrix tries to follow, more so than an RFC. Most likely browsers are more in line with the RFC these days, but have we tested that?

ato:
> Have we looked at performance implications of doing toArray() frequently?

Here's the 4.3.6 HttpClient code:

```java
      final List<Cookie> cookies = new ArrayList<Cookie>(cookieStore.getCookies());
      ...
      for (final Cookie cookie : cookies) {
```

Here's the 4.5.7 HttpClient code:

```java
        final List<Cookie> cookies = cookieStore.getCookies();
        ...
        for (final Cookie cookie : cookies) {
```

`new ArrayList` calls `toArray()` internally, and the new code calls `toArray()` in `iterator()` so they seem roughly equivalent, a single array copy is made either way. So while the array copy could possibly now be avoided in the new version of HttpClient it doesn't seem to be a performance regression.

> Is the subdomain thing the only implication of switching from CookieSpecs.BROWSER_COMPATIBILITY to CookieSpecs.DEFAULT?

It appears that in 4.5.7 BROWSER_COMPATIBILITY is now an alias for DEFAULT and the old behaviour is not available:

https://github.com/aache/httpcomponents-client/blob/4.5.7/httpclient/src/main/java/org/apache/http/impl/client/CookieSpecRegistries.java#L56-L58

There are many changes between 4.3's compat mode and 4.5.7's default mode.

> browser behavior is the real spec heritrix tries to follow, more so than an RFC. Most likely browsers are more in line with the RFC these days, but have we tested that?

So what do browsers do? Apparently a lot of different things (both to the RFC and to each other): https://inikulin.github.io/cookie-compat/

The subdomain thing appears to be test DOMAIN029 with Chrome, Firefox and Safari being RFC6265 compliant on that particular issue but IE and Edge not.

So browsers are not by any means RFC6265-strict but neither is CookieSpecs.DEFAULT (the RFC6265 strict mode is CookieSpecs.STANDARD_STRICT ).

DEFAULT is documented as:
```
    /**
     * The default policy. This policy provides a higher degree of compatibility
     * with common cookie management of popular HTTP agents for non-standard
     * (Netscape style) cookies.
     */
    public static final String DEFAULT = "default";
```

I guess someone could rig up a harness to run HttpClient against that same set of tests and see how close it is to Chrome as presumably the standard to aim for at the moment by virtue of popularity.

nlevitt:
Thanks @ato! Good enough for me!

PR #255, opened 2019-04-18
:wrench: upgrade spring dependencies from 3.0.5-RELEASE to 4.3.21.REL…

### LINKED ISSUE #254: Upgrade dependencies to spring 4.x.x
Heritrix-commons depends on spring 3.0.5-RELEASE which is an outdated version.

The use of heritrix-commons as a library may lead to dependencies issues.

jkowalczyk:
…EASE (latest)

Closes #254

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -137,27 +137,22 @@
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-core</artifactId>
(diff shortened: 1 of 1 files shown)


jkowalczyk:
There is no test on the spring package. So I don't know if this huge upgrade can break anything.

As far as I know, the spring schemas defined in the job example file (profile-crawler-beans.cxml), are already in version 3.0. So it may be without impact

PR #257, opened 2019-04-30
WIP: ExtractorYoutubeDL

nlevitt:


contrib/src/main/java/org/archive/modules/extractor/ExtractorYoutubeDL.java
```diff
@@ -0,0 +1,342 @@
...
+            proc = pb.start();
+        } catch (IOException e) {
+            logger.log(Level.WARNING, "youtube-dl failed " + pb.command(), e);
+            return (List<JsonObject>) Collections.EMPTY_LIST;
+        }
+
+        String stdout = null;
+        String stderr = null;
+        try {
+            stdout = readToEnd(
```
ato: Just a heads up I've been bitten by this pattern hanging. If the child process writes more than the OS' pipe buffer size then the child's write syscalls will block. Meanwhile the parent Java process is waiting reading stdout and never progresses to reading stderr. As far as I know the only safe way to separately read both stdout and stderr in Java is to do it from separate threads or to redirect one of them to a temp file. :-(

nlevitt: Thank you for pointing this out! I've been bitten by this myself but I guess I didn't learn my lesson.

Issue #250, opened 2019-03-29
Support full wildcard syntax in robots.txt directives


anjackson:
We only support trailing `*` wildcards at present.  Ideally we should support wildcards as defined in https://developers.google.com/search/reference/robots_txt 

The code to modify would be:

https://github.com/internetarchive/heritrix3/blob/05811705ed996122bea1f4e034c1ed5f7a07240f/modules/src/main/java/org/archive/modules/net/RobotsDirectives.java#L40-L42

The actual wildcards are not that difficult, but getting the precedence right is harder.  Perhaps we can use a standard library e.g. the crawler commons code?

anjackson:
see https://github.com/crawler-commons/crawler-commons/blob/bef1b8437e63930bcbab82a3f754bf835cda5cca/src/main/java/crawlercommons/robots/SimpleRobotRules.java#L153

jr-ewing:
Maybe interesting https://github.com/google/robotstxt

ato:
There's a Java port of Google's parser too https://github.com/google/robotstxt-java/ but unfortunately it doesn't seem to be in Maven Central.

PR #262, opened 2019-07-05
Support for extracting URLs in sitemaps

kris-sigur:
This builds on work by @anjackson . I'm submitting it for discussion. It requires more testing before considering merging.

I've modified the two extractors he wrote, eliminating the BL specific, recrawl, elements. I renamed them to match Heritrix extractor naming conventions. I also added a new hop type, M (Manifest) for the links extracted by them. Other than that, they are mostly as Andy wrote them, barring a few minor tweaks.

As the sitemap extractor requires a new dependency (crawler-commons), I added that to the 'modules' sub-project. This creates a dependency conflict as crawler-commons requires a newer version of commons-io. I upgraded the commons-io (from 1.4 to 2.4) in the 'commons' sub-project. As far as I can tell this doesn't cause any regression.


anjackson:
BTW, taking on `crawler-commons` as a dependency also means we could consider allowing use of (or switching to) their [robots.txt parser](http://crawler-commons.github.io/crawler-commons/1.0/crawlercommons/robots/SimpleRobotRulesParser.html). It's more powerful that the current H3 one, and so should be able to resolve #250.

modules/src/main/java/org/archive/modules/extractor/ExtractorRobotsTxt.java
```diff
@@ -0,0 +1,115 @@
...
+        try {
+
+            // Clone the CrawlURI and change hop path and avoid queueing
+            // sitemaps as prerequisites (i.e. strip P from hop path).
+            CrawlURI curiClone = new CrawlURI(curi.getUURI(),
+                    curi.getPathFromSeed().replace("P", ""), curi.getVia(),
+                    curi.getViaContext());
+
+            // Also copy the source over:
+            curiClone.setSourceTag(curi.getSourceTag());
```
anjackson: Oops, came across a bug. You must only copy over the `SourceTag` if it's not `null`, otherwise other code breaks. This is because `getData().hasKey(x)` is used as a test instead of checking for `null` value.

```
            if (curi.getSourceTag() != null) {
                curiClone.setSourceTag(curi.getSourceTag());
            }
```

anjackson:
Hm, looks like this needs some more work before we can go ahead, so I'll defer this for now.

anjackson:
Okay, this looks like a sensible port of the UKWA code with the revisit logic taken out. I'll look at shifting our modules to be subclasses of these later on.

PR #280, opened 2019-08-12
Mitigate random CookieStore.testConcurrentLoad test failures

### LINKED ISSUE #274: CookieStoreTest.testConcurrentLoad fails randomly
Just hit this with a Travis on #271. It looks like someone else also hit when building from source in #175.

```
testConcurrentLoad(org.archive.modules.fetcher.CookieStoreTest)  Time elapsed: 1.188 sec  <<< FAILURE!
junit.framework.AssertionFailedError
	at junit.framework.Assert.fail(Assert.java:47)
	at junit.framework.Assert.assertTrue(Assert.java:20)
	at junit.framework.Assert.assertTrue(Assert.java:27)
	at org.archive.modules.fetcher.CookieStoreTest.testConcurrentLoad(CookieStoreTest.java:391)
```

The problem seems to have been introduced by #133:

> I don't think we need the synchronization in fact. Because the 50 cookie limit doesn't have to be completely strict. It's just a cutoff to avoid problems, and if a domain has 51 cookies, or even 60, it's unlikely to be a problem. So my suggestion is to remove the synchronization, and change CookieStoreTest:350 to something like `assertTrue(domainCounts.get(domain) <= BdbCookieStore.MAX_COOKIES_FOR_DOMAIN + 5);`, i.e. check that it's not way over the limit.

<strike>Unfortunately I don't think this assumption is safe. Not just because it can cause the tests to randomly fail either. At the NLA we've actually hit a gotcha in several different applications where Java's HashMap goes into an infinite loop when accessed concurrently without synchronization.

https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6423457
> If a HashMap is used in a concurrent setting with insufficient synchronization,
> it is possible for the data structure to get corrupted in such a way that
> infinite loops appear in the data structure and thus get() could loop forever.</strike>

Edit: Confused myself reading the code. The HashMap is not accessed in multiple threads.

ato:
The arbitary value `25` was used but in practice it's quite possible
for more than 25 writing threads to have checked the cookie count
limit before adding their cookie. In practice we see Travis failing
on this test quite often, every few builds in fact.

~I think using `threads.length` (i.e. 200) should cover the worst
case possibility where every thread reads a stale count and tries
to add their cookie.~ Edit: Updated to remove the test entirely.

Fixes #274

### DIFF
--- a/modules/src/test/java/org/archive/modules/fetcher/CookieStoreTest.java
+++ b/modules/src/test/java/org/archive/modules/fetcher/CookieStoreTest.java
@@ -388,7 +388,9 @@ public class CookieStoreTest extends TmpDirTestCase {
         }
         
         for (String domain: domainCounts.keySet()) {
-            assertTrue(domainCounts.get(domain) <= BdbCookieStore.MAX_COOKIES_FOR_DOMAIN + 25);
+            // the cookie store intentionally doesn't synchronize so we allow up to thread.length
+            // additional cookies over the limit
+            assertTrue(domainCounts.get(domain) <= BdbCookieStore.MAX_COOKIES_FOR_DOMAIN + threads.length);
         }        
     }
     
--- a/modules/src/test/java/org/archive/modules/fetcher/CookieStoreTest.java
+++ b/modules/src/test/java/org/archive/modules/fetcher/CookieStoreTest.java
@@ -338,62 +338,7 @@ public class CookieStoreTest extends TmpDirTestCase {
         assertTrue(bdbCookieList.size() > 3000);
         assertCookieListsEquivalent(bdbCookieList, basicCookieStore().getCookies());        
     }
-    
-    public void testConcurrentLoad() throws IOException, InterruptedException {
-        bdbCookieStore().clear();
-        basicCookieStore().clear();
-        final Random rand = new Random();
 
-        Runnable runnable = new Runnable() {
-            @Override
-            public void run() {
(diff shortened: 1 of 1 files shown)


anjackson:
The change looks fine, but I'm still a bit confused as to what this test is actually _testing_. If we make the upper-bound so large it can never possibly be exceeded, what's the point of the test?

nlevitt:
> The change looks fine, but I'm still a bit confused as to what this test is actually testing. If we make the upper-bound so large it can never possibly be exceeded, what's the point of the test?

I'm thinking you're right @anjackson. Maybe we should just drop the test. The assumption when we wrote the test was that a race condition would not be so frequent in practice. We've seen that under the contrived conditions created by the test case, it is frequent. But that's ok

ato:
Andy writes:

> Isn't this failure just showing that MAX_COOKIES_FOR_DOMAIN gets ignored?

My understanding is it doesn't get ignored, it just means under high thread contention it can be exceeded by up to the thread count, as each thread re-entering will recheck the count. I can't say I'm 100% confident of that understanding though.

Under normal configuration I don't think multiple threads will be writing to the same domain anyway, although I wouldn't rely on that were the side effects any worse.

Noah writes:

> Maybe we should just drop the test.

I'm all for that. Updated to do so.

nlevitt:
Thanks guys!

Issue #277, opened 2019-08-08
Checkpoints 'spoiled' when used to resume crawls


anjackson:
We've seen occasional mysterious issues when resuming crawls from checkpoints multiple times.  Close inspection of the behaviour when resuming the crawl indicates that checkpoints can only reliably be used once.

The code uses the [DbBackup](https://docs.oracle.com/cd/E17277_02/html/java/com/sleepycat/je/util/DbBackup.html) helper to manage backups. This depends on calling [startBackup](https://docs.oracle.com/cd/E17277_02/html/java/com/sleepycat/je/util/DbBackup.html#startBackup--), the docs for which note:

> Following this method call, all new data will be written to other, new log files. In other words, the last file in the backup set will not be modified after this method returns.

i.e. the BDB-JE makes sure the backup is consistent and after that those files should no longer be altered (BDB is an append-only DB system). The [backup documentation](https://docs.oracle.com/cd/E17277_02/html/GettingStartedGuide/backuprestore.html) implies this flushed, sync'd consistency is necessary for the backup to work.

(Note that a H3 checkpoint is not same thing as a [BDB-JE checkpoint](https://docs.oracle.com/cd/E17277_02/html/java/com/sleepycat/je/Environment.html#checkpoint-com.sleepycat.je.CheckpointConfig-) - the former is a point in time backup, but the latter is a flush/sync operation).

However, when resuming from a crawl, rather than copying the checkpoint files ([as recommended by the documentation](https://docs.oracle.com/cd/E17277_02/html/GettingStartedGuide/catastrophicrecovery.html)), Heritrix3 uses hard links (and cleans out any other state files not part of the checkpoint).

This causes an issues because, having resumed a crawl, I noticed that the last `.jdb` file in the checkpoint was being changed! From the backup behaviour, we might expect that existing files would not be changed, but in fact when resuming a crawl, the system proceeds by appending data to the last `.jdb` file.

As this file is a hard link, this activity also changes the contents of the checkpoint.  Furthermore, if we are resuming the crawl from one older checkpoint among many, all subsequent checkpoints are also modified.

As an example, we recently attempted to resume from a checkpoint, hit some difficulties and had to force-kill the crawler. After this, we attempted to re-start from that checkpoint, and hit some very strange behaviour that hung the crawler. (Fortunately, we happen to have a backup of those checkpoints!)

[...]

PR #281, opened 2019-08-12
WIP: Upgrade BDB JE to version 7.5.11

anjackson:
For #277 and other issues relating to using a very old version of BDB JE, this patch updates to the latest version. Apart from one test which relied on `Thread.interrupt()` (which makes BDB flake out), the upgrade appears to be straightforward.

However, marking this as a Work In Progress until I get a chance to run a large-scale test with it.

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -34,7 +34,7 @@
 		<dependency>
 			<groupId>com.sleepycat</groupId>
 			<artifactId>je</artifactId>
-			<version>4.1.6</version>
+			<version>7.5.11</version>
 		</dependency>
 		<dependency>
 			<groupId>commons-lang</groupId>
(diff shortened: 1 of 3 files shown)


anjackson:
Small scale testing works fine and indicates that the `.jdb` file re-use problem in #277 is resolved by this change.

anjackson:
Having tested this in production, it seems to work fine. The only issue was that that on-disk format has changed since 4.1.6. See https://docs.oracle.com/cd/E17277_02/html/changelog.html

Specifically, 4.1.6 databases require at least one manual step (see `DbPreUpgrade_4_1` part of the changelog.  I found this difficult to execute in concert with Heritrix3's checkpoint management.

So, the question is, do we want to try to support upgrading existing crawler state databases, or just say this is a new release means you need to wipe your state?

ato:
We discussed this on the OH-SOS call and nobody present had a need to preserve state between crawler upgrades. It sounds reasonable to me that an upgrade will invalidate checkpoints. I guess if someone really needs to resume a pre-upgrade crawl they can still use the recover log right?

anjackson:
That seems reasonable to me.

Have we heard from anyone from IA?

nlevitt:
Hey, sorry. Andy brought this to my attention again on slack. Archive-It is coming up with a plan. I will alert other IA folks using heritrix.

anjackson:
Okay, so I think we're good to go ahead here.

PR #290, opened 2019-12-10
Crawltrap regex timeout

csrster:
At first sight this might seem like an odd thing to want to do. Everybody uses Java regex matching all the time and people are mostly quite happy with it. But it turns out that the Java regex engine has no guaranteed upper-bound on the time taken to match and we have in practice run into cases where the runtime has exploded exponentially.

Typically this is the result of a pathologically dumb regex meeting a crazy url (I have provided an example!) - which is why having a timeout is a good idea because it is then possible to log the precise set of circumstances which led to the timeout and diagnose the underlying pathology.

The fix is coded so that the default behaviour is backwards compatible with previous versions.


csrster:
The travis build failure comes from adding a Test class with no tests (because the one test in it is commented out). Probably the best thing to do is just to rename the test so it is recognised as a real test by junit. (The test is not great because the behaviour on timeout is the same as the behaviour if the regex match completes - ie both cases are treated as non-matching. So you can't prove that you're testing what you want to test unless you inspect the logging output - always a bad idea.)

modules/src/main/java/org/archive/modules/deciderules/MatchesListRegexDecideRule.java
```diff
@@ -91,7 +107,17 @@ protected boolean evaluate(CrawlURI uri) {
...
 
         for (Pattern p: regexes) {
-            boolean matches = p.matcher(str).matches();
+            boolean matches = false;
+            if (getTimeoutPerRegexSeconds() <= 0) {
+                matches = p.matcher(str).matches();
+            } else {
+                CompletableFuture<Boolean> matchesFuture = CompletableFuture.supplyAsync(() -> p.matcher(str).matches());
+                try {
+                    matches = matchesFuture.get(getTimeoutPerRegexSeconds(), TimeUnit.SECONDS);
```
ato: My understanding is that CompletableFuture.get() with a timeout doesn't actually interrupt the long-running regex, it simply doesn't wait for it to finish before progressing. Therefore the bad regex will continue running in the background on the common ForkJoin thread pool. After the MatchRule is executed a few times all threads in the pool will be busy processing bad regexs and all newly executed regex match rules will fail with a timeout thus corrupting the result of even good regexs.

Amusingly the StackOverflow answer for the general regex-with-a-timeout problem is actually answered by @kris-sigur where he suggests Heritrix's InterruptibleCharSequence class. https://stackoverflow.com/a/910798

Unfortunately we can't simply wrap str in InterruptibleCharSequence and call `matchesFuture.cancel(true)` as CompleteFuture's cancel() is [documented](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#cancel-boolean-) as not actually interrupting:

> mayInterruptIfRunning - this value has no effect in this implementation because interrupts are not used to control processing.

Using FutureTask instead like this seems to work though:

```java
        InterruptibleCharSequence interruptible = new InterruptibleCharSequence(str);
        FutureTask<Boolean> matchesFuture = new FutureTask<>(() -> p.matcher(interruptible).matches());
        ForkJoinPool.commonPool().submit(matchesFuture);
        try {
            matchesFuture.get(getTimeoutPerRegexSeconds(), TimeUnit.SECONDS);
        } catch (TimeoutException e) {
            matchesFuture.cancel(true);
            logger.warn("Timed out after " + getTimeoutPerRegexSeconds() + " seconds waiting for '" + p + "' to match.');
        }
```

Another approach that is a little more code but may be cleaner because it doesn't use extra threads at all is also suggested in the next SO answer:

https://stackoverflow.com/a/11348374

csrster: I prefer the FutureTask method using standard API calls rather than roll-your-own multi-threading.

csrster: I have now updated the branch with the requested changes. - Colin

ato:
Thanks! :)

PR #292, opened 2020-01-14
Replace A_TIMESTAMP with A_FETCH_BEGAN_TIME

hennekey:
It seems that A_TIMESTAMP went out of favor quite a long time ago.
A_FETCH_BEGAN_TIME is used within FetchHistoryProcessor and throws an
exception as is because of it.

### DIFF
--- a/contrib/src/main/java/org/archive/modules/recrawl/FetchHistoryHelper.java
+++ b/contrib/src/main/java/org/archive/modules/recrawl/FetchHistoryHelper.java
@@ -27,6 +27,7 @@ import java.util.Map;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.archive.modules.CoreAttributeConstants;
 import org.archive.modules.CrawlURI;
 
 /**
@@ -37,10 +38,6 @@ import org.archive.modules.CrawlURI;
  */
 public class FetchHistoryHelper {
   private static final Log logger = LogFactory.getLog(FetchHistoryHelper.class);
-  /**
-   * key for storing timestamp in crawl history map.
-   */
-  public static final String A_TIMESTAMP = ".ts";
 
   /**
    * returns a Map to store recrawl data, positioned properly in CrawlURI's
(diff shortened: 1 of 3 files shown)


ato:
This seems reasonable but I'm not familiar enough with this module to confidently review this. I'm hoping someone more knowledgeable about this comments, otherwise I'll merge it in a few days if we don't hear anything.

nlevitt:
This is @kngenie's code. I don't know if it's still used.

hennekey:
I was trying to use it in conjunction with the [HBase recrawl modules](https://github.com/internetarchive/heritrix3/tree/master/contrib/src/main/java/org/archive/modules/recrawl/hbase) but everything besides these class seems to use `A_FETCH_BEGAN_TIME` and so they're writing to the wrong attribute and `FetchHistoryProcessor` throws an exception here: https://git.hubteam.com/HubSpot/heritrix3/blob/51c7c55e4ac928f9d9c5dbd5cfb830ad0b7fd767/modules/src/main/java/org/archive/modules/recrawl/FetchHistoryProcessor.java#L124

anjackson:
We use `A_TIMESTAMP` but can switch if that makes sense. However, I'm having trouble seeing what exactly you are referring to @hennekey because your links go to your private GitHub Enterprise instance. Any chance you can open it up? Or refer to this repo instead?

hennekey:
@anjackson Sorry about that, I've updated the link to the code.

Any explanation about how you make use of `A_TIMESTAMP` would be helpful so I can do the right thing.

anjackson:
Thanks!

We use it in much the same way, I think. We use it for [deduplication](https://github.com/ukwa/ukwa-heritrix/blob/56fc4ecd03882c54f180e44409551f8ed82cf96b/src/main/java/uk/bl/wap/modules/recrawl/OutbackCDXPersistLoadProcessor.java#L123) and for [checking how recently we visited to see if we need to re-crawl yet](https://github.com/ukwa/ukwa-heritrix/blob/11bd1133de07bfe03c9ded47909a4d2087140dcd/src/main/java/uk/bl/wap/modules/deciderules/RecentlySeenDecideRule.java#L329). But it looks like I hit similar issues because [I populate both fields](https://github.com/ukwa/ukwa-heritrix/blob/0d09a51e494322bee3e3c686ddd093c72081366b/src/main/java/uk/bl/wap/util/OutbackCDXClient.java#L364-L365).

I'm not sure how important these semantics are -- the timestamp in question may not necessarily be the time the fetch began, but in practice I think it always is, so I think it's fine to reduce it all to that name.

hennekey:
Yeah, I did reference that comment.
It occurs to me now that perhaps semantically `A_TIMESTAMP` is meant to be the time the history map entry was created for that CrawlURI and `A_FETCH_BEGAN_TIME` is more specifically the point in time that the fetch action started.
I defer to you to the semantics; you've obviously got more experience with it. I'd be happy to make whatever changes we deem necessary as a result.

hennekey:
Your use of this constant also makes it clear that it's a more dangerous change than I thought. It is `public` afterall so there may be other users.

Perhaps the proper solution is to clarify the semantics and use the proper value in `FetchHistoryProcessor` instead?

PR #295, opened 2020-01-15
Disable test that connects to wwwb-dedup.us.archive.org

ato:
http://wwwb-dedup.us.archive.org:8083/web/timemap/cdx?url=http%3A%2F%2Fwww.mext.go.jp%2Fnull.gif&limit=-1
is returning 503 intermittently and so this test is breaking the Travis builds of
unrelated changes.


hennekey:
e.g. https://travis-ci.org/internetarchive/heritrix3/builds/637102680

PR #296, opened 2020-01-15
Set JUnit version to latest

hennekey:
Updating JUnit to begin updating the unit tests

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -119,7 +119,6 @@
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
-			<version>3.8.2</version>
 			<scope>compile</scope>
 		</dependency>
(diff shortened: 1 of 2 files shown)


ato:
I'm no maven expert but my (possible wrong) understanding is setting the version in dependencyManagement only works if junit is referred to in a normal dependencies section without a version tag. Since commons/pom.xml pulls in junit as a compile-time dependency with an explicit version I don't think this patch actually has any effect?

hennekey:
You're right about commons. I can fix that.
I think setting the version in `dependencyManagement` in the parent POM is a good idea because all the children can make use of that to be on the same version. My motivation for this is to allow me to update/add unit tests along the way, and I'm no longer familiar with JUnit 3.

ato:
Thanks!

PR #304, opened 2020-01-24
Replace custom Base32 encoding

hennekey:
Guava is available so a custom implementation is unnecessary.

### DIFF
--- a/commons/src/main/java/org/archive/util/Base32.java
+++ b/commons/src/main/java/org/archive/util/Base32.java
@@ -18,142 +18,25 @@
  */
 package org.archive.util;
 
+import com.google.common.io.BaseEncoding;
+
 /**
- * Base32 - encodes and decodes RFC3548 Base32
- * (see http://www.faqs.org/rfcs/rfc3548.html )
- *
- * Imported public-domain code of Bitzi.
- *
- * @author Robert Kaye
- * @author Gordon Mohr
+ * @deprecated Use {@link com.google.common.io.BaseEncoding#base32()}
  */
+@Deprecated
 public class Base32 {
-    private static final String base32Chars =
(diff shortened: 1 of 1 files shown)


commons/src/main/java/org/archive/util/Base32.java
```diff
@@ -18,142 +18,25 @@
...
-                digit = (currByte >> (8 - (index + 5))) & 0x1F;
-                index = (index + 5) % 8;
-                if (index == 0)
-                    i++;
-            }
-            base32.append(base32Chars.charAt(digit));
-        }
-
-        return base32.toString();
+      return BaseEncoding.base32().encode(bytes);
```
ato: There seems to be an output change here:

original: `Base32.encode("a") => "ME"`
new: `Base32.encode("a") => "ME======"`

We might need `omitPadding()` to ensure this is an exact drop in replacement.

commons/src/main/java/org/archive/util/Base32.java
```diff
@@ -18,142 +18,25 @@
...
-            if (b < 0) {
-                b += 256;
-            }
-            System.out.print((Integer.toHexString(b + 256)).substring(1));
-        }
-        System.out.println();
-        System.out.println("Reencoded: " + Base32.encode(decoded));
+    @Deprecated
+    static public byte[] decode(final String base32)  {
+      return BaseEncoding.base32().decode(base32);
```
ato: This is throwing an exception on lowercase input such as in `ContentDigestHistoryTest.testBasics()`.

hennekey: Got it.

ato:
Thanks!

PR #306, opened 2020-01-24
Fix stream closed exception by not closing output stream

ato:
ServerCall.writeResponseBody() flushes it after we return so it must
remain open.

Fixes #305

Issue #305, opened 2020-01-24
exception logged when opening/saving crawler-beans.cxml via web interface editor


ldko:
Building from [master](https://github.com/internetarchive/heritrix3/tree/9927e1325f132adc4e31d8bf7a990f3c3a928004) and running in either Java 8 or 11, when I use the Heritrix web interface to open or save the crawler-beans.cxml file, editing and saving functionality seems to be working okay, but I am seeing errors like the following in the heritrix_out.log:
```
2020-01-24 21:48:19.935 SEVERE thread-26 org.restlet.engine.adapter.ServerAdapter.commit() An exception occurred writing the response entity
java.io.IOException: Stream closed
        at java.base/sun.nio.cs.StreamEncoder.ensureOpen(StreamEncoder.java:45)
        at java.base/sun.nio.cs.StreamEncoder.flush(StreamEncoder.java:152)
        at java.base/java.io.OutputStreamWriter.flush(OutputStreamWriter.java:254)
        at org.restlet.representation.CharacterRepresentation.write(CharacterRepresentation.java:68)
        at org.restlet.engine.adapter.ServerCall.writeResponseBody(ServerCall.java:507)
        at org.restlet.engine.adapter.ServerCall.sendResponse(ServerCall.java:451)
        at org.restlet.ext.jetty.internal.JettyServerCall.sendResponse(JettyServerCall.java:309)
        at org.restlet.engine.adapter.ServerAdapter.commit(ServerAdapter.java:191)
        at org.restlet.engine.adapter.HttpServerHelper.handle(HttpServerHelper.java:144)
        at org.restlet.ext.jetty.JettyServerHelper$WrappedServer.handle(JettyServerHelper.java:237)
        at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:370)
        at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:267)
        at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:305)
        at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:103)
        at org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.onFillable(SslConnection.java:427)
        at org.eclipse.jetty.io.ssl.SslConnection.onFillable(SslConnection.java:321)
        at org.eclipse.jetty.io.ssl.SslConnection$2.succeeded(SslConnection.java:159)
[...]

PR #308, opened 2020-01-30
Fix stream closed exception for Paged view

ldko:
Related to #305 the stream closed exception is also showing up when clicking "more" to get the "Paged view" of the Job Log or Crawl Log for a job in the Heritrix web user interface. This commits a similar fix to what is seen in #306.


ato:
Thanks!

PR #312, opened 2020-02-13
Exclude hbase-client's guava 12 transitive dependency

### LINKED ISSUE #311: Contrib project has a maven dependency with an older version of guava library.
Changes introduced in https://github.com/internetarchive/heritrix3/pull/304 broke the contrib project.

The `org.archive.modules.recrawl.hbase` package has a resolved dependency on version 12 of guava and other versions of the hbase client I've sampled in the POM resolve to guava v12 or lower.

```xml
<dependency>
        <groupId>org.apache.hbase</groupId>
        <artifactId>hbase-client</artifactId>
        <version>0.98.6-cdh5.3.5</version>
        <exclusions>
                <exclusion>
                        <artifactId>jets3t</artifactId>
                        <groupId>net.java.dev.jets3t</groupId>
                </exclusion>
                <exclusion>
                        <artifactId>junit</artifactId>
                        <groupId>junit</groupId>
                </exclusion>
                <!-- tools.jar is not available in JDK 11 so exclude it
                     hbase-client accidentally leaked it as a transitive dependency
                     https://issues.apache.org/jira/browse/HBASE-13963 --> 
                <exclusion>
                        <groupId>jdk.tools</groupId>
                        <artifactId>jdk.tools</artifactId>
                </exclusion>
        </exclusions>
</dependency>
```

I also realize there are no tests for this so it didn't show up in CI or anywhere else.

ato:
Guava 12 from hbase-client is closer to the root of the dependency tree than guava 17 from webarchive-commons so Maven prefers it. But recent changes to heritrix-commons rely on classes in the newer version of Guava. So let's ensure webarchive-commons wins.

Hopefully this doesn't break the hbase module, I have no way of testing it.

Fixes #311

### DIFF
--- a/contrib/pom.xml
+++ b/contrib/pom.xml
@@ -33,6 +33,11 @@
 					<groupId>jdk.tools</groupId>
 					<artifactId>jdk.tools</artifactId>
 				</exclusion>
+				<!-- prefer guava 17 from webarchive-commons over guava 12 from hbase-client -->
+				<exclusion>
+					<groupId>com.google.guava</groupId>
+					<artifactId>guava</artifactId>
+				</exclusion>
 			</exclusions>
 		</dependency>
 		<dependency>


hennekey:
It might be better to set the versions via dependency management at a top level POM to ensure they're consistent across the modules.

ato:
It's currently not a direct dependency of any heritrix modules. Although I guess now that were using it directly perhaps it should be.

BitBaron:
We're looking at dropping Hbase support all together. We may not need it any more.

ato:
I'm going to merge this now as a stopgap fix until discussion over the future of the hbase recrawl module in #313 is resolved. If we still keep getting reports of problems then I think we should roll back the Base32 change.

PR #313, opened 2020-02-13
Remove Old Hbase and Kafka Classes

BitBaron:


hennekey:
I'm using the HBase functionality for managing recrawling of URLs.

Why is it being removed?

BitBaron:
It was being dropped b/c we thought we were the only ones using those bits. And we don't need them any more.

Is https://github.com/internetarchive/heritrix3/pull/312 working for you?

hennekey:
I haven't experienced any issues with it yet.

anjackson:
Hm, we're using Kafka but it looks like I made a [new variation](https://github.com/ukwa/ukwa-heritrix/blob/da03f65611d4a9ea2a73f5f837ca69da3b5e5ae2/src/main/java/uk/bl/wap/crawler/postprocessor/KafkaKeyedCrawlLogFeed.java) rather than using the `KafkaCrawlLogFeed` directly.

PR #314, opened 2020-03-04
Utilize the `d` parameter

hennekey:
There are uses of this class that necessitate a lower false positive
rate than the value that was hard-coded. By making use of the parameter,
as indicated in the JavaDoc, the resulting delegate should meet the
demands of the caller.

### DIFF
--- a/commons/src/main/java/org/archive/util/BloomFilter64bit.java
+++ b/commons/src/main/java/org/archive/util/BloomFilter64bit.java
@@ -80,7 +80,7 @@ public class BloomFilter64bit implements Serializable, BloomFilter {
      * @param roundUp if true, round bit size up to next-nearest-power-of-2
      */
     public BloomFilter64bit(final long n, final int d, Random weightsGenerator, boolean roundUp ) {
-        delegate = com.google.common.hash.BloomFilter.create(Funnels.unencodedCharsFunnel(), Ints.saturatedCast(n), 0.0000003);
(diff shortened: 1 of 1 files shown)


commons/src/main/java/org/archive/util/BloomFilter64bit.java
```diff
@@ -80,7 +80,7 @@ public BloomFilter64bit( final long n, final int d, boolean roundUp) {
      * @param roundUp if true, round bit size up to next-nearest-power-of-2
      */
     public BloomFilter64bit(final long n, final int d, Random weightsGenerator, boolean roundUp ) {
```
hennekey: I'd love to remove the unused `Random` parameter, but I'm afraid it'd break clients unknown. What do you think about deprecating this constructor, as well as the methods that are seemingly unnecessary and used only in the tests (e.g. https://github.com/internetarchive/heritrix3/blob/master/commons/src/main/java/org/archive/util/BloomFilter64bit.java#L157 & https://github.com/internetarchive/heritrix3/blob/master/commons/src/test/java/org/archive/util/BloomFilterTest.java#L159) to further clean this up?

anjackson: Not a bad idea, but I'm not sure if the deprecation warnings show up when you're tying things together with Spring?

anjackson: Well, either way, yes, please mark that as deprecated.

anjackson: okay fine let's take this as-is then.

Issue #319, opened 2020-04-01
Add support for the SFTP protocol


bnfleb:
Allow the writing of sftp:// URLs by Heritrix

ato:
Closing this now that #320 has been merged.

PR #320, opened 2020-04-01
Add support for the SFTP protocol

bnfleb:
Support for SFTP protocol (issue #319)
I create a new client and a new fetcher to support SFTP.

### DIFF
--- a/commons/src/main/java/org/archive/net/ClientSFTP.java
+++ b/commons/src/main/java/org/archive/net/ClientSFTP.java
@@ -22,9 +22,9 @@ public class ClientSFTP {
 	protected StringBuilder controlConversation;
 
 	protected Socket dataSocket;
-	protected static Session session = null;
-	protected static Channel channel = null;
-	protected static ChannelSftp channelSFTP = null;
+	protected Session session = null;
+	protected Channel channel = null;
+	protected ChannelSftp channelSFTP = null;
 
 	public ClientSFTP() {
 		this.controlConversation = new StringBuilder();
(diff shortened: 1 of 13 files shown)


commons/src/main/java/org/archive/net/ClientSFTP.java
```diff
@@ -0,0 +1,153 @@
...
+		this.logger.info("channelSftp put: " + paramString);
+		channelSFTP.put(paramFileInputStream, paramString);
+		return true;
+	}
+
+	protected boolean downRemoteSingleFile(String paramString1, String paramString2) throws Exception {
+		FileOutputStream fileOutputStream = new FileOutputStream(paramString2);
+		channelSFTP.get(paramString1, fileOutputStream);
+
+		this.logger.info("channelSftp download file: " + paramString1);
```
ato, line 145: Let's give these parameters meaningful names. I can overlook it when there's just one but `paramString2` is too much. ;-)

```suggestion
	protected boolean downRemoteSingleFile(String src, String dest) throws Exception {
		FileOutputStream fileOutputStream = new FileOutputStream(dest);
		channelSFTP.get(src, fileOutputStream);

		this.logger.info("channelSftp download file: " + src);
```

ato (changes requested):
Sorry for taking so long to do this review.

This mostly looks okay but I noticed two serious problems (static session/channel etc fields in ClientSFTP and a byte loop index) and also have a couple of minor suggestions to improve the code readability.

bnfleb:
Sorry for the old code and meaningless variable names, it came from a jar of heritrix1 but adapted for heritrix3.
It seems that the static fields are not necessary.

ato:
Thanks!

PR #323, opened 2020-04-10
Add parsing for HTML tags (data-*)

clawia:
Add parsing for data-src, data-srcset, data-original and data-original-set tags to original ExtractorHTML.

### DIFF
--- a/modules/src/main/java/org/archive/modules/extractor/ExtractorHTML.java
+++ b/modules/src/main/java/org/archive/modules/extractor/ExtractorHTML.java
@@ -28,10 +28,8 @@ import java.util.Iterator;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


modules/src/main/java/org/archive/modules/extractor/ExtractorHTML.java
```diff
@@ -43,1000 +45,999 @@
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import com.google.common.base.Strings;
+
+import au.id.jericho.lib.html.Element;
```
ato, line 50: I think this patch needs to be cleaned up a bit before it can be merged. It seems to be modifying nearly every line in the file? It's also introducing a bunch of unused imports? I haven't reviewed the actual code changes as it's too hard to see what changed. :(

clawia:
@ato I have just pushed new version of ExtractorHTML with only my changes, not a formatting classe.
I add overrides in the JerochoExtractorHTMLTest for my news tests.
Is better for you?

PR #326, opened 2020-04-17
Add real crawlStatus in the crawlReport

clawia:
During the crawl, we have "crawl status: Finished - Abnormal exit from crawling" in the crawlSummary report.
I have changed the crawlStatus to have the same info as the job main page:
- crawl status: Pausing - Active
- crawl status: Paused - Active
- crawl status: Running - Active
- crawl status: Empty - Active
- crawl status: Stopping - Active
when isRunning()=true.

The crawl summary report is always re generated to have the correct crawlStatus.


anjackson:
Hmm, seeing some build errors on Travis that look legit:

```
[ERROR] /home/travis/build/internetarchive/heritrix3/engine/src/main/java/org/archive/crawler/reporting/CrawlSummaryReport.java:[40,31] cannot find symbol
  symbol:   variable StringUtils
  location: class org.archive.crawler.reporting.CrawlSummaryReport
```

clawia:
My bad! I don't push the import. It should be better.

ato:
This looks good to me. Since the previous behavior was nonsensical while the crawl was active I think the risk of breaking anyone's existing workflow should be relatively low too.

PR #328, opened 2020-04-30
Fix match result is always false in MatchesListRegexDecideRule

morokosi:
It looks like #290 introduced a bug that `MatchesListRegexDecideRule#evaluate()` always returns false because future result is discarded.
This fix respects the returned future value.
Thanks,


ato:
Thanks!

PR #334, opened 2020-05-20
Warc convention for storing ftp responses has been to use a WARC reso…

adam-miller:
…urce record and not a response record. Changing to be consisten with the default WARCWriterProcessor


ato:
Sorry @adam-miller, I merged this thinking it was merging into master but realised after that it's the master-ait-contrib branch.

PR #348, opened 2020-08-04
Fixes leaky file handles

adam-miller:


anjackson:
I've only been able to give this a quick scan, but it looks good. Thanks @adam-miller

PR #366, opened 2021-02-15
Upgrade to Spring 5.3.3

### LINKED ISSUE #254: Upgrade dependencies to spring 4.x.x

### LINKED ISSUE #337: Spring instanciation broken for MatchesListRegexDecideRule
In [this context](https://github.com/internetarchive/heritrix3/pull/290#discussion_r366711640) a lambda function was introduced to [MatchesListRegexDecideRule](https://github.com/internetarchive/heritrix3/blame/ef38dc08c357bcb2353bbfbcd00a2c3370aad9c7/modules/src/main/java/org/archive/modules/deciderules/MatchesListRegexDecideRule.java#L117).

This class no longer works when created as part of the crawler beans -- instead producing [this inscrutable error](https://stackoverflow.com/a/46963786/6689).

As far as I can tell, this should not work unless we upgrade Spring, so I'm rather confused as to how this came about.

ato:
Spring 3.0.5 is 10 years old now and is incompatible with classes using Java 9+ bytecode and even with Java 8 classes that use lambdas (see #337). We remove spring-asm as it is now part of spring-core module.

This introduces some deprecation warnings as Spring now recommends constructor injection instead of the [`@Required`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/annotation/Required.html) annotation. I believe that would require a breaking change in the job config format (`<constructor-arg>` instead of `<property>`) though so I'm leaving the deprecated annotations in place for now.

@kris-sigur and I have both tested Heritrix against Spring 5.3.3 and neither of us encountered any compatibility issues.

Closes #254. Closes #255. Fixes #337.

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -136,27 +136,22 @@
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-core</artifactId>
-			<version>3.0.5.RELEASE</version>
+			<version>${spring.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-beans</artifactId>
-			<version>3.0.5.RELEASE</version>
+			<version>${spring.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-context</artifactId>
-			<version>3.0.5.RELEASE</version>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework</groupId>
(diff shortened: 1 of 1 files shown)


kris-sigur:
Given the variety of configurations (plus possibly people running with their own custom modules), it is very possible that this breaks *something* even if @ato and I didn't find any issues. But I think it is necessary to take this step soon if Heritrix development is to continue.

PR #397, opened 2021-06-14
Upgrade httpclient to 4.5

anjackson:
This is a fresh attempt to upgrade HTTP Client to version 4.5.x, for #245


anjackson:
Okay, so I've now implemented cookie expiration, while updating to httpclient 4.5.x. If you're able to kick the tyres on this one @kris-sigur that'd be great!

anjackson:
In case it helps, the latest `SNAPSHOT` release is based off this 4.5.x fork: https://oss.sonatype.org/content/repositories/snapshots/org/archive/heritrix/heritrix/3.4.0-SNAPSHOT/ 

e.g. https://oss.sonatype.org/content/repositories/snapshots/org/archive/heritrix/heritrix/3.4.0-SNAPSHOT/heritrix-3.4.0-20210617.135449-15-dist.zip

modules/src/main/java/org/archive/modules/fetcher/AbstractCookieStore.java
```diff
@@ -221,7 +221,21 @@ protected LimitedCookieStoreFacade(List<Cookie> cookies) {
...
-            throw new RuntimeException("not implemented");
+            int expiredCount = 0;
+            for( Cookie c : cookies) {
+                boolean expired = AbstractCookieStore.this.expireCookie(c, date);
+                if( expired ) {
+                    logger.fine("Expired cookie: " + c + " for date: " + date);
+                    expiredCount++;
+                }
+            }
+            // Replace the cookie list, if any expired:
```
ato, line 232: I think this comment might be a stray comment leftover from the earlier commit? I don't think we replace the list any more?

modules/src/main/java/org/archive/modules/fetcher/BdbCookieStore.java
```diff
@@ -78,7 +79,7 @@
         @Override public int size() { return wrapped.size(); }
         @Override public boolean isEmpty() { throw new RuntimeException("not implemented"); }
         @Override public boolean contains(Object o) { throw new RuntimeException("not implemented"); }
-        @Override public Iterator<T> iterator() { throw new RuntimeException("not implemented"); }
+        @Override public Iterator<T> iterator() { return (Iterator<T>) Arrays.asList(wrapped.toArray()).iterator(); }
```
ato, line 82: We should probably also update the class-level javadoc that says iterator() is not implemented, because now it is. :-)

modules/src/main/java/org/archive/modules/fetcher/BdbCookieStore.java
```diff
@@ -143,6 +144,22 @@ public void addCookieImpl(Cookie cookie) {
...
         }
     }
+    
+    public boolean expireCookie(Cookie cookie, Date date) {
+        byte[] key;
+        try {
+            key = sortableKey(cookie).getBytes("UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            throw new RuntimeException(e); // impossible
+        }
```
ato, line 154: I think if we use java.nio.charset.StandardCharsets.UTF_8 we don't need to try/catch:

```suggestion
        byte[] key = sortableKey(cookie).getBytes(StandardCharsets.UTF_8);
```

modules/src/main/java/org/archive/modules/fetcher/BdbCookieStore.java
anjackson: Not quite sure how to put it, but I'll have a go!

modules/src/main/java/org/archive/modules/fetcher/AbstractCookieStore.java
anjackson: Yep, earlier iteration under there. I'll drop the comment.

ato (approved):

anjackson:
As this is a fairly invasive change, I'm thinking I'd like @kris-sigur to take a look at this before we merge?

anjackson:
I think the only remaining problem is the dropping of the `AuthCache` - this is apparently used to avoid any initial communiction with the server to determine the authentication needed (see [this StackOverflow answer](https://stackoverflow.com/a/21592593/6689)).  i.e. without this, there should be a `401` transaction followed by the 'real' transaction.

Right now I'm unsure as to whether this will cause problems, or if (in effect) H3 never sees that initial negociation and only records the successful transaction.

ato:
I tested both versions with basic auth (I don't have a server handy which supports digest auth). With the new code there is as expected an increase in the number of requests sent to the server but no change in the number of WARC records written.

Crawl configured with CredentialStore and two seeds that have HTTP basic auth enforced by the server: `/auth.html` and `/auth.html?2`.

With the old code credentials seem to be cached between requests as only the first request generates a 401:
```
# nginx acces.log (old code)
127.0.0.1 - - [05/Jul/2021:15:23:47 +0900] "GET /auth.html HTTP/1.0" 401 179 "-" "Mozilla/5.0 (compatible; heritrix/3.4.0-SNAPSHOT-2021-07-05T06:21:14Z +http://example.org/)"
127.0.0.1 - user [05/Jul/2021:15:23:47 +0900] "GET /auth.html HTTP/1.0" 200 40 "-" "Mozilla/5.0 (compatible; heritrix/3.4.0-SNAPSHOT-2021-07-05T06:21:14Z +http://example.org/)"
127.0.0.1 - user [05/Jul/2021:15:23:47 +0900] "GET /auth.html?2 HTTP/1.0" 200 40 "-" "Mozilla/5.0 (compatible; heritrix/3.4.0-SNAPSHOT-2021-07-05T06:21:14Z +http://example.org/)"
```

With hew new code they aren't and the second seed also generates a 401:
```
# nginx acces.log (new code)
127.0.0.1 - - [05/Jul/2021:15:24:29 +0900] "GET /auth.html HTTP/1.0" 401 179 "-" "Mozilla/5.0 (compatible; heritrix/3.4.0-SNAPSHOT-2021-07-05T06:24:14Z +http://example.org/)"
127.0.0.1 - user [05/Jul/2021:15:24:29 +0900] "GET /auth.html HTTP/1.0" 200 40 "-" "Mozilla/5.0 (compatible; heritrix/3.4.0-SNAPSHOT-2021-07-05T06:24:14Z +http://example.org/)"
127.0.0.1 - - [05/Jul/2021:15:24:29 +0900] "GET /auth.html?2 HTTP/1.0" 401 179 "-" "Mozilla/5.0 (compatible; heritrix/3.4.0-SNAPSHOT-2021-07-05T06:24:14Z +http://example.org/)"
127.0.0.1 - user [05/Jul/2021:15:24:29 +0900] "GET /auth.html?2 HTTP/1.0" 200 40 "-" "Mozilla/5.0 (compatible; heritrix/3.4.0-SNAPSHOT-2021-07-05T06:24:14Z +http://example.org/)"
```

Only the first 401 is recorded in the WARC though so the new httpclient actually writes the same number of records as the old httpclient:
```
# warc output (new code)
$ jwarc cdx jobs/test-auth/latest/warcs/WEB-20210705062428220-00000-672759~jib.localdomain~8444.warc.gz 
http://localhost/robots.txt 20210705062428 http://localhost/robots.txt text/plain 200 WWUS2CRACUBU6QT3GVCQLRNJ4YF7TNCI - - 439 711 jobs/test-auth/latest/warcs/WEB-20210705062428220-00000-672759~jib.localdomain~8444.warc.gz
http://localhost/auth.html 20210705062429 http://localhost/auth.html text/html 401 RJIRBTXAKQULLMFM6DG5R7TJBXU2LDG6 - - 450 1901 jobs/test-auth/latest/warcs/WEB-20210705062428220-00000-672759~jib.localdomain~8444.warc.gz
http://localhost/auth.html 20210705062429 http://localhost/auth.html text/html 200 CVE3L55SJCUGAJN2GF5UH4GB2CU26IQS - - 424 3068 jobs/test-auth/latest/warcs/WEB-20210705062428220-00000-672759~jib.localdomain~8444.warc.gz
http://localhost/auth.html?2 20210705062429 http://localhost/auth.html?2 text/html 200 CVE3L55SJCUGAJN2GF5UH4GB2CU26IQS - - 427 4233 jobs/test-auth/latest/warcs/WEB-20210705062428220-00000-672759~jib.localdomain~8444.warc.gz
http://localhost/favicon.ico 20210705062430 http://localhost/favicon.ico text/html 404 47UIAI63XRRUNU2U77U7W7NZK6EIYIUZ - - 412 5405 jobs/test-auth/latest/warcs/WEB-20210705062428220-00000-672759~jib.localdomain~8444.warc.gz

```

The 401 responses never appears in crawl.log with either version:
```
$ cat jobs/test-auth/latest/logs/crawl.log 
2021-07-05T06:24:28.225Z     1         44 dns:localhost P http://localhost/auth.html text/dns #004 20210705062427697+13 sha1:QQRA3RZDT3JHH3C5ZMGTPX6IBO2JPVQI - -
2021-07-05T06:24:28.685Z   200         36 http://localhost/robots.txt P http://localhost/auth.html text/plain #019 20210705062428446+171 sha1:WWUS2CRACUBU6QT3GVCQLRNJ4YF7TNCI - -
2021-07-05T06:24:29.714Z   200         40 http://localhost/auth.html - - text/html #019 20210705062429699+6 sha1:CVE3L55SJCUGAJN2GF5UH4GB2CU26IQS - 4t
2021-07-05T06:24:29.949Z   200         40 http://localhost/auth.html?2 - - text/html #021 20210705062429943+3 sha1:CVE3L55SJCUGAJN2GF5UH4GB2CU26IQS - -
2021-07-05T06:24:30.199Z   404        153 http://localhost/favicon.ico PI http://localhost/robots.txt text/html #014 20210705062430194+2 sha1:47UIAI63XRRUNU2U77U7W7NZK6EIYIUZ - -
```

anjackson:
Well, constantly renegotiating the auth on each access doesn't seem like desirable behaviour. Re-enabling the cache would avoid this, for BASIC auth, but break things (or at least break the tests) for DIGEST auth.

I don't really understand what's changed between HttpClient versions, I guess. It seems like the `FetchHTTPRequest` uses a fresh HttpClientContext for every request, so I don't understand how it would ever have been able to track the `nc` across requests! 

Whatever happened, it seems HttpClient no longer supports caching DIGEST auth (if I'm reading this correctly): https://issues.apache.org/jira/browse/HTTPCLIENT-1855

So, least worst option seems to be to support caching but only for BASIC.

anjackson:
Unless I broke something somewhere else, this should now use the `AuthCache` except for `DigestScheme`.  As @ato has shown, the crawler should still create the right WARCs, but will make a `401` connection every time when using `DigestScheme`.

anjackson:
Okay I believe this is fine so I'm going to go ahead and merge it.

PR #400, opened 2021-06-18
Remove unnecessary fiddling with VIA path in ExtractorRobotsTxt

kris-sigur:
When extracting sitemap URIs from robots.txt files, effort is made to clip the "P" (referring to the robots.txt being determined as a prerequisite) from the VIA path. Consequently, the sitemap URIs discovered from seeds will just start with "M" as if they are derived from a seed, but still with the robots.txt as parent in the crawl log.

The explanation for this behavior given in the code comments may be based on a misunderstanding or faulty code elsewhere. The VIA path is not used to determine if a crawlUri is a prerequisite (there is a field on crawlUri dedicated for that) and even if some module is sniffing the VIA path, they should only consider the last character. If they are looking at any other character it is a bug in that module and needs to be fixed there.

This PR removes this logic and handles the extracted sitemap URI normally, with the robots.txt as parent. I've verified that (running with the usual default configuration at least) sitemaps are not treated as prerequisites and are scheduled like any other discovered URI (i.e. same as L hops).

@anjackson

### DIFF
--- a/modules/src/main/java/org/archive/modules/extractor/ExtractorRobotsTxt.java
+++ b/modules/src/main/java/org/archive/modules/extractor/ExtractorRobotsTxt.java
@@ -54,18 +54,6 @@ public class ExtractorRobotsTxt extends ContentExtractor {
     @Override
     protected boolean innerExtract(CrawlURI curi) {
         try {
-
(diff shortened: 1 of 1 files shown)


anjackson:
Ah fair enough 😬 😄

PR #402, opened 2021-06-20
Remove dependency on mg4j

ato:
This eliminates one of the jars that Heritrix depends on that isn't in Maven Central.

There were only two usages. The first usage was a method in a CrawlerJournal which as far as I can tell is unused and can be removed.

The second usage was an unnecessary use of MutableString as a line buffer when saving cookies to a file. Since there doesn't seem any particular reason to buffer per-line here let's just use a BufferedWriter instead on the whole file which is also slightly simpler and more efficient.


anjackson:
Oof good call, that's quite a big dependency for a small/questionable optimisation.  Patch looks good to me.

anjackson (commented):
LGTM

PR #403, opened 2021-06-21
Add ExtractorChrome to contrib

ato:
This adds a link extractor that uses a headless browser via the Chrome
Devtools Protocol. It works with Chromium and Google Chrome and
while untested likely also other Blink-based browsers such as Brave and
Microsoft Edge. It also works with Firefox which implements a subset of
the protocol.

In addition to ExtractorChrome itself this adds a couple of classes
under org.archive.net.chrome for communicating with the browser:

* ChromeProcess - manages the lifecycle of the browser child process
* ChromeClient  - client for the Chrome Devtools Protocol
* ChromeWindow  - client for controlling a particular browser window

Why this extractor when we already have Umbra? The primary goal is to
have something that works out of the box without any additional
infrastructure beyond having a browser installed. A longer term goal is
to be able to take advantage of tighter integration between Heritrix
and the browser to do things like recording browser subrequests to WARC
files or running ExtractorHTML against the loaded DOM.

While this extractor is already usable it is currently quite bare bones
and only extracts links from `<a>` and `<area>` tags. The intent is to
expand it with more functionality over time with features like
intercepting subrequests, reusing previously saved resources, taking
screenshots or running JavaScript behaviours.

### DIFF
--- a/contrib/pom.xml
+++ b/contrib/pom.xml
@@ -83,6 +83,11 @@
 			<artifactId>gson</artifactId>
 			<version>2.8.6</version>
 		</dependency>
+		<dependency>
+			<groupId>org.java-websocket</groupId>
+			<artifactId>Java-WebSocket</artifactId>
+			<version>1.5.2</version>
+		</dependency>
 	</dependencies>
 	<repositories>
 		<repository>
--- a/contrib/src/main/java/org/archive/modules/extractor/ExtractorChrome.java
+++ b/contrib/src/main/java/org/archive/modules/extractor/ExtractorChrome.java
@@ -1,0 +1,206 @@
+/*
+ *  This file is part of the Heritrix web crawler (crawler.archive.org).
+ *
+ *  Licensed to the Internet Archive (IA) by one or more individual
+ *  contributors.
+ *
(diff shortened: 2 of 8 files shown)


anjackson:
I'm not feeling well-placed to review this in detail. It looks about right, but I'm not really familiar enough with this kind of direct protocol interaction to say. And I'm not really in a position to test it at the moment. That said, given it's a new `contrib` I don't think that should block this merge.

ato:
There's some disagreement about the best way forward for this. The most popular alternative idea seems to be to bundle an external Node.js-based companion process (either playwright-driver or a derivative of browsertrix-crawler) to interface with the browser directly instead of using CDP.

If others want to follow up on such an alternative they're more than welcome to but I don't plan to do so myself because my personal motivation for exploring this instead of the existing Umbra integration was to have something that can be used out of the box without external dependencies other than browser itself.

CC @ikreymer

ato:
I'm going provisionally merge this so that others can play with it. I still like this approach and I'm willing to maintain it for now. It can be removed in future again if it bitrots, proves not to be useful or if one of the various alternatives that have been suggested gets implemented and becomes popular.

PR #409, opened 2021-06-25
Fix WARC-IP-Address and use a common server-ip CrawlURI attribute for all protocols

### LINKED ISSUE #396: Add WARC-IP-Address header to WARCWriterChainProcessor
The original `WARCWriterProcessor` added a `WARC-IP-Address` header if the IP of the host is known:

https://github.com/internetarchive/heritrix3/blob/37ce8d694590b0cf8cbe0a38a58c5f8ee719c4f0/modules/src/main/java/org/archive/modules/writer/WARCWriterProcessor.java#L278

The newer `WARCWriterChainProcessor` uses a builder for HTTP responses, but this does not include the IP address:

https://github.com/internetarchive/heritrix3/blob/37ce8d694590b0cf8cbe0a38a58c5f8ee719c4f0/modules/src/main/java/org/archive/modules/warc/HttpResponseRecordBuilder.java#L35

Other builders do, like this:

https://github.com/internetarchive/heritrix3/blob/37ce8d694590b0cf8cbe0a38a58c5f8ee719c4f0/modules/src/main/java/org/archive/modules/warc/DnsResponseRecordBuilder.java#L42-L45

It's not clear the original complexity is needed:

https://github.com/internetarchive/heritrix3/blob/37ce8d694590b0cf8cbe0a38a58c5f8ee719c4f0/modules/src/main/java/org/archive/modules/writer/WriterPoolProcessor.java#L378-L401

But if the IP is handy, it should be included in the HTTP response.

ato:
This fixes #396 by introducing a new protocol-agnostic server-ip attribute on CrawlURI. The HTTP, FTP, WHOIS and DNS fetch processors now write to it and the appropriate WARC record builders copy from it to populate WARC-IP-Address.

The old A_DNS_SERVER_IP_LABEL and A_WHOIS_SERVER_IP attributes are fully replaced by the new generic one.

The ServerCache lookup is kept as a fallback in IpAddressSetDecideRule.getHostAddress(curi) since unlike the WARC writer it needs to know the IP address before the URI has been fetched.

### DIFF
--- a/modules/pom.xml
+++ b/modules/pom.xml
@@ -62,6 +62,12 @@
 			<version>1.6.6</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.ftpserver</groupId>
+			<artifactId>ftpserver-core</artifactId>
+			<version>1.1.1</version>
+			<scope>test</scope>
+		</dependency>
     <dependency>
       <groupId>com.github.crawler-commons</groupId>
       <artifactId>crawler-commons</artifactId>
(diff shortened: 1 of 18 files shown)


anjackson (approved):
Looks right and covered both the 'old' and 'new' WARC writer systems.

PR #415, opened 2021-07-19
Merge ait-qa branch to re-syncronised development

anjackson:
This is a merge intended to re-sync with the `ait-qa` branch.


ato:
OK by me. Although this does include the removal of the hbase module which someone objected to last year (see #313).

anjackson:
Okay,  I think that's synced up, but with HBase modules restored from the master branch.

PR #433, opened 2021-09-01
Access builds.archive.org over HTTPS by using the Wayback Machine

ato:
Newer versions of Maven block access to non-HTTPS repositories in order
to provide protection against man in the middle attacks (see #310).
Unfortunately builds.archive.org does not provide public HTTPS access.

However the Wayback Machine is available over HTTPS and has the needed
artifacts (javaswf, kryo, minlog, reflectasm) archived. This is a little
oddly circular but solves the immediate problem.

I changed the order of the repositories and added central explicitly so
that Maven will only fallback to Wayback for the artifacts that aren't
in the first two.

I also turned off snapshots. It looks like builds.archive.org has
stopped generating new Heritrix snapshots anyway - I don't have
access to the IA Jenkins to see why but based on the timing I'd guess it
might have been building with Java 7.


ato:
On second thought maybe this isn't such a good idea due to the risk of malicious content being injected by deposited WARCs. I'm not sure how easy it would be to pull off but global wayback data gets sourced from a lot of different places.

Other ideas I had were:
* Host a repository on Github Pages: e.g. https://github.com/iipc/heritrix-dependencies
* Push the needed artifacts to Central under a different groupId. (Not sure if this is possible or not.)

PR #435, opened 2021-09-10
Adding fixed versions and moving to Zulu builds of OpenJDK

geertjanw:
It is good practice to do CI/CD against the latest version, e.g., 11, as well as a fixed version, e.g., 11.0.3, so that when the build fails on the latest version but passes on the fixed version, you'll know the problem is with the latest version and not with your source code. However, Adopt is no more, Temurin and Zulu are the two valid options, though Temurin is new and doesn't provide all the versions, while Zulu is more complete and can therefore support this solution.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -20,7 +20,7 @@ jobs:
       uses: actions/setup-java@v2
       with:
         java-version: ${{ matrix.jdk }}
-        distribution: 'temurin'
+        distribution: 'zulu'
     - name: Cache local Maven repository
       uses: actions/cache@v2
       with:
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
-        jdk: [8, 11, 16]
+        jdk: [8, 8.0.192, 11, 11.0.3, 16, 16.0.2]
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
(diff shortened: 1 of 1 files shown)


ato:
Thanks for the heads up about AdoptOpenJDK's rebranding to Temurin.  I don't recall seeing your name in connection with the web archiving community before, so welcome. I see that you work for Azul, the creators of Zulu, and have submitted several similar pull requests to other open source projects.

I must admit to some confusion about the stated motivation for switching to Zulu. If the reason for adding fixed versions to the build matrix is to be a reference to distinguish incompatibilities introduced by newer JDK versions versus source changes then surely we could pick any known good fixed version as the reference? If so then could we not just pick fixed versions that Temurin actually has builds for instead of versions that only Zulu has?

geertjanw:
Sure, though which fixed versions are those? Azul supports Temurin as well as Zulu, either one or both are fine to use.

geertjanw:
From the point of choosing between Temurin and Zulu, at this stage, it is clear that Zulu provides all OpenJDK versions, major and minor, while for Temurin, which has only very recently been created, history begins at that point. For your users, who are most likely using JDK versions that Temurin does not support, it would make most sense to make sure the code works with older rather than only the latest JDK versions.

ato:
I've applied 7f7d1cda8763a5a601c1abc13b1b416b0e14d3c7 and also updated the AdoptOpenJDK link in the docs as that solves the immediate issue of the [name change](https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/).

The fixed versions idea to differentiate the type of regressions is interesting but it hasn't been a problem to date and I don't personally want to encourage or support the use of unpatched JVMs so I think I'll leave it there for now unless our other maintainers feel differently.

PR #441, opened 2021-10-04
Enabled configurable url-matching and extraction for sitemaps.

csrster:
I've added two parameters to ExtractorSitemap that make it possible to find sitemaps and their urls more leniently. I've written them to be backward-compatible with the current implementation - ie omitting the two new fields in the extractor bean causes a fallback to the current behaviour. The two new parameters are

1. urlPattern: used to identify sitemaps. Currently the code looks at the mime-type and sniffs the start of the file. Not all sites specify the mime-type of sitemaps correctly so this provides an alternative way to identify them.
2. enableLenientExtraction. When true, the code extracts all urls found in the sitemap, ignoring the scoping rules of the sitemap specification. Not all websites conform to the spec.

### DIFF
--- a/modules/src/main/java/org/archive/modules/extractor/ExtractorSitemap.java
+++ b/modules/src/main/java/org/archive/modules/extractor/ExtractorSitemap.java
@@ -30,6 +30,19 @@ public class ExtractorSitemap extends ContentExtractor {
     private static final Logger LOGGER = Logger
             .getLogger(ExtractorSitemap.class.getName());
 
+    /**
+     * If urlPattern is not null then any url marked as a sitemap and matching the pattern is
+     * assumed to be a sitemap. Otherwise the mime-type is checked (must be "text/xml" or "application/xml") and the
+     * file is "sniffed" for the expected start of a sitemap file.
+     */
+    private String urlPattern = null;
+
+    /**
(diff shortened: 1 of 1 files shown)


modules/src/main/java/org/archive/modules/extractor/ExtractorSitemap.java
```diff
@@ -181,4 +197,32 @@ private void recordOutlink(CrawlURI curi, URL newUri, Date lastModified,
...
+    }
+
+    /**
+     * If urlPattern is not null then any url marked as a sitemap and matching the pattern is
+     * assumed to be a sitemap. Otherwise the mime-type is checked (must be "text/xml" or "application/xml") and the
+     * file is "sniffed" for the expected start of a sitemap file.
+     * @param urlPattern the url pattern to match
+     */
+    public void setUrlPattern(String urlPattern) {
+        this.urlPattern = urlPattern;
```
ato, line 211: Maybe we should compile the regular expression here (`Pattern.compile(urlPattern)`) or just take a Pattern object instead of a String? Other places in the config like MatchesRegexDecideRule accept a Pattern object. That way if someone makes a syntax error in the pattern it'll be caught immediately when the job is built instead of later in the crawl when they first encounter a sitemap.

PR #454, opened 2021-12-17
Added support for SOCKS5 proxy systems, ARM Linux support

hyl:
Hi all,

This PR adds support for SOCKS5 proxying. The host and port details can be configured within the fetchHttp bean in `crawler-beans.cxml`:

```
<bean class="org.archive.modules.fetcher.FetchHTTP" id="fetchHttp">
    <!--  ... -->
    <property name="socksProxyHost" value="127.0.0.1"/>
    <property name="socksProxyPort" value="24000"/>
</bean>
```

In addition, we've updated the `pom.xml` files to allow symlinking on ARM Linux and to provide an alternate repository source for some dependencies that were failing from `download.oracle.com,maven` in `commons`.

I've opened this as a draft PR for the following reasons:

1. To check what additional documentation is required for the new property values, and where I should add it;
2. To confirm that my testing approach is appropriate: I wasn't sure if we should move the sample SOCKS5 server into `FetchHTTPTest` rather than `FetchHTTPTests` so thought I'd throw it open to the community for consensus.

I hope this work is useful - any questions, give me a shout and more than happy to answer.


commons/pom.xml
```diff
@@ -16,6 +16,10 @@
 	</description>
 
 	<repositories>
+		<repository>
+			<id>builds.archive.org,maven2</id>
+			<url>https://builds.archive.org/maven2</url>
+		</repository>
```
ato, line 22: I think builds.archive.org isn't available publicly over https?
```sh
$ curl https://builds.archive.org/maven2
curl: (7) Failed to connect to builds.archive.org port 443 after 154 ms: Connection refused
```

modules/pom.xml
```diff
@@ -79,11 +79,17 @@
...
+			<groupId>com.github.bbottema</groupId>
+			<artifactId>java-socks-proxy-server</artifactId>
+			<version>1.1.0</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+		  <groupId>com.github.crawler-commons</groupId>
+		  <artifactId>crawler-commons</artifactId>
+		  <version>1.0</version>
+		</dependency>
```
ato, line 92: I don't think this change actually needs crawler-commons?

modules/src/main/java/org/archive/modules/fetcher/FetchHTTPRequest.java
```diff
@@ -168,6 +171,12 @@ public ServerCacheResolver(ServerCache serverCache) {
     protected HttpHost targetHost;
     protected boolean addedCredentials;
     protected HttpHost proxyHost;
+
+    // SOCKS5 proxy configuration
+    protected String socksProxyHost;
+    protected Integer socksProxyPort;
+    protected Boolean useSocksProxy;
```
ato, line 178:
```suggestion
    protected boolean useSocksProxy;
```
The constructor always sets this and the null case doesn't seem meaningful so maybe this could be a primitive?

ato (commented):
Nice! This functionality would be handy. I can see us making use of it in future too.

> To check what additional documentation is required for the new property values, and where I should add it;

It might be a good idea to note in the setSocksProxyHost method doc that setting a SOCKS proxy will override any set HTTP proxy. Otherwise I think this is fine.

I guess it might be nice if we had a general section about configuring proxy servers in the [configuration guide](https://heritrix.readthedocs.io/en/latest/configuring-jobs.html) (docs/configuring-jobs.rst) but I'd be happy to merge this without that and the new settings will at least appear in the bean reference and javadoc.

> To confirm that my testing approach is appropriate: I wasn't sure if we should move the sample SOCKS5 server into FetchHTTPTest rather than FetchHTTPTests so thought I'd throw it open to the community for consensus.

The comment at the top of FetchHTTPTests gives the impression that it was split up we don't have to start and stop the test web servers for each test. Perhaps junit at the time this was written didn't have a way to do setup/teardown code at the class level (like `@AfterClass` with JUnit 4 or `@AfterAll` with JUnit 5) and so they had to do it at the TestSuite level. Since the SOCKS server is only used by one test though there's no need to keep it running for the whole suite so the way you've done it seems fine to me.

hyl:
Hi @ato ,

So sorry for the delay in responding to your feedback.

I've addressed the comments below in the commits this morning and added a new SOCKS5 section to the documentation: I'll mark the PR as ready for review :)

ato:
Thanks. Merged as dd01689acea06ac401dfa84546639a20e7718304 with some trivial cleanups. I dropped the commit that updates JNA for ARM as we've removed the need for JNA entirely (#468).

After merging the SOCKS test failed on GitHub CI but I couldn't reproduce the failure locally. Reading the code I theorized since the SOCKS server binds its socket in a background thread there's a race where the client can connect before the server is listening. I added a workaround in d79649199585921916119977c1f27683537c870f that seems to have fixed it.

Issue #191, opened 2017-10-30
"RIS already open for ToeThread..." exception during https pages crawl over proxy


WI-IT:
When I try to crawl https pages over a proxy with Heritrix 3, I get following exceptions:

`java.io.IOException: RIS already open for ToeThread #5: https://www.XXX/robots.txt
	at org.archive.io.RecordingInputStream.open(RecordingInputStream.java:84)
	at org.archive.util.Recorder.inputWrap(Recorder.java:185)
	at org.archive.modules.fetcher.FetchHTTPRequest$RecordingHttpClientConnection.getSocketInputStream(FetchHTTPRequest.java:648)
	at org.apache.http.impl.BHttpConnectionBase.ensureOpen(BHttpConnectionBase.java:131)
	at org.apache.http.impl.DefaultBHttpClientConnection.sendRequestHeader(DefaultBHttpClientConnection.java:140)
	at org.apache.http.protocol.HttpRequestExecutor.doSendRequest(HttpRequestExecutor.java:203)
	at org.apache.http.protocol.HttpRequestExecutor.execute(HttpRequestExecutor.java:121)
	at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:254)
	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:195)
	at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:86)
	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:184)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:72)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:57)
	at org.archive.modules.fetcher.FetchHTTPRequest.execute(FetchHTTPRequest.java:751)
	at org.archive.modules.fetcher.FetchHTTP.innerProcess(FetchHTTP.java:658)
	at org.archive.modules.Processor.innerProcessResult(Processor.java:175)
	at org.archive.modules.Processor.process(Processor.java:142)
	at org.archive.modules.ProcessorChain.process(ProcessorChain.java:138)
	at org.archive.crawler.framework.ToeThread.run(ToeThread.java:148)`

marhop:
[...]

PR #457, opened 2022-01-10
Fix issue#191: "RIS already open for ToeThread..." exception during https pages crawl over proxy

ClemensRobbenhaar:
See [issue#191](https://github.com/internetarchive/heritrix3/issues/191) for the problem description.

### Description of the Change

When using a proxy, HTTPS-Requests are handled in two steps:

a) the client sends a request of the form:

    CONNECT <hostname>:<port> HTTP/1.1

  to the proxy. The proxy then opens an encrypted connection to
  the actual server and passes this to the client.

b) the client sends the data over the secure connection
   as plain HTTP-request like

    GET /some/path/name?param1=value1 HTTP/1.1

   and receives the response in plain, too (as it is tunneled via an encrypted connection)

To support this behavior, the current code in the `FetchHTTPRequest` is changed in two places:

a) in the constructor around line 189 do not use the full URI as data to be send after the verb
   if the method is HTTPS. Instead send the same line as without proxy

b) in the methods for the inner class `RecordingHttpClientConnection` check if we have an https request and a proxy in use. In that case do not wrap the first created input/output stream, but only the ones afterwards.

Here it would be better to check if the socket is the one used for the `CONNECT` or the one used to tunnel the actual request/response data (as the `XXX` comments suggests, which are already present in the code).
 However this information is not available inside this class, so instead make the assumption that the input / output streams for the `CONNECT` are created first, and the one for the data to be recorded afterwards.

This avoids trying to use the same RecordingInput/OutputStream twice without the need to remove the safety checks in these classes which prevents erroneous reuse.   

### Testing

There is one new test `testHttpsProxy` in the `FetchHTTPTests`, but I am not sure if that is sufficiently complete.

Additionally I tested the changes manually using Apache HTTPD as a proxy, running on the same host as the Herirtix crawler.

The following snippet was used as configuration file `/etc/apache2/sites-enabled/proxy-vhost.conf`:

```
Listen 8000

<VirtualHost *:8000>
  ProxyRequests On

  ProxyVia On

  SSLProxyVerify none
  SSLProxyCheckPeerCN off
  SSLProxyCheckPeerName off

  <Proxy "*">
    Require ip 127.0.0 192.168
  </Proxy>

  LogFormat "%h %l %u %t \"%r\" %>s %b [%D] \"%{User-agent}i\"" withduration
  CustomLog ${APACHE_LOG_DIR}/proxy_access.log withduration

</VirtualHost>
```

Enabled modules:

- `mod_proxy`
- `mod_ssl`
- `mod_proxy_http`
- `mod_proxy_connect`

The the `fetchHttp` bean in the heritirix config has the proxy configured in the usual way:

     <property name="httpProxyHost" value="127.0.0.1" />
     <property name="httpProxyPort" value="8000" />

The HTTPS-Server used as target for the crawl was also an apache httpd; as it turned out this server is not affected by the first change (it seems not to mind if you send it requests which looks like proxy requests). The second change was necessary to crawl any HTTPS-URIs via proxy at all and did not affect the crawling of HTTP-URIs.

If you have change requests, suggestions for improvements or any other questions/comments about the patch, please let me know.

### DIFF
--- a/modules/src/main/java/org/archive/modules/fetcher/FetchHTTPRequest.java
+++ b/modules/src/main/java/org/archive/modules/fetcher/FetchHTTPRequest.java
@@ -189,7 +189,12 @@ public class FetchHTTPRequest {
         if (StringUtils.isNotEmpty(proxyHostname) && proxyPort != null) {
             this.proxyHost = new HttpHost(proxyHostname, proxyPort);
             this.requestConfigBuilder.setProxy(this.proxyHost);
-            requestLineUri = curi.getUURI().toString();
+            if ("https".equalsIgnoreCase(curi.getUURI().getScheme())) {
+                // with SSL connections the hostname is already send with the CONNECT
+                requestLineUri = curi.getUURI().getEscapedPathQuery();
+            } else {
+                requestLineUri = curi.getUURI().toString();
+            }
         } else {
             requestLineUri = curi.getUURI().getEscapedPathQuery();
         }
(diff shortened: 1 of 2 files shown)


ato:
Thanks!

PR #461, opened 2022-01-28
Add conf to not allow TLDs as seeds found via redirect from other seeds

kris-sigur:
By default this does not change existing behavior. Adds the ability to exclude as possible seeds, any URL whose host is a top level domain (as determined by the ArchiveUtils method isTld from the webarchive-commons dependency).

This is useful where scope is determined by seeds as a single one of these can open up an entire TLD, which is rarely wanted.

Requested for NetarchiveSuite. They've tested this and found it resolves their issue.


clawia:
From tests run at BnF
1) Crawl with seeds:
http://allmyfaves.fr
http://www.allmyfaves.fr

and **seedsRedirectNewSeedsAllowTLDs=false**

**H3 stayed on http://allmyfaves.fr domain** even if there was a redirection http://allmyfaves.fr/ -> https://fr/ . 

2) Crawl with seeds: 
http://allmyfaves.fr
http://www.allmyfaves.fr

and **seedsRedirectNewSeedsAllowTLDs=true**

Same H3 behaviour with the current release: **https://fr/ is added as a new seed**.

Issue #211, opened 2018-08-02
Do not require DNS when using a web proxy


marhop:
Hi,

I am trying to crawl some WWW domain from behind a web proxy in a corporate network using Heritrix build 3.3.0-20180727.011238-114 without success. Heritrix just hangs right after the crawl is started. I suppose this is caused by Heritrix taking a rather unusual approach to DNS queries when using a web proxy. Let me explain:

The DNS servers in our corporate network only resolve host names from our local network. (And I cannot use external DNS servers because of firewall rules.) That's OK because all external requests are routed through a web proxy anyway. A client tells the web proxy the (external) URL it wishes to access, and the web proxy takes care of everything, including DNS resolution (which is done by forwarding the request to a parent proxy that does the actual DNS resolution using other, "more knowledgable" DNS servers).

Long story short, when using a web proxy, it's not necessary to query the DNS. Most tools do it this way, and "just work". (Example: curl makes DNS requests only when *not* using a web proxy.)

However, Heritrix *always* makes DNS requests, regardless of its proxy configuration. If a DNS request fails, it does not even go on asking the proxy for the URL it tries to crawl, although the proxy could easily fetch the content. Instead, it just hangs.

So may I suggest that Heritrix should *not* query the DNS when it is configured to use a web proxy? Or if it does, that at least it should continue asking the web proxy even if the DNS request fails?

Probably related: #198

Thanks,
Martin

ato:
Hi Martin,

Heritrix does its own DNS lookups as it writes the DNS records and IP addresses to the WARC files. Other features like geolocation and ip address decide rules also depend on knowing the IP addresses. Ignoring hostnames that do not resolve while perhaps not essential likely also helps keep a certain amount of garbage URLs out of the queues early.

If you don't need those features and would like to try to modify Heritrix to work without DNS, a quick and dirty workaround might be a new implementation of org.archive.modules.net.ServerCache which returns a dummy IP address for every lookup. A proper solution would need an option to disable the DNS pecondition check in PreconditionEnforcer and to modify the WARC and ARC writers to work without IP addresses. Right now they assume IP addresses are always available.

Cheers,

Alex

marhop:
Hi Alex,

thanks for the quick and thorough explanation, greatly appreciated! Should I read your second paragraph more like "we're looking at it" or "we definitely won't do it ourselves, but wouldn't reject a decent pull request either"?

Thanks,
Martin

ato:
[...]

PR #476, opened 2022-04-08
Issue211: support dns over https if local DNS is not working / available

ClemensRobbenhaar:
As per comment: https://github.com/internetarchive/heritrix3/issues/211#issuecomment-409925550 and in the hope that this PR is considered "decent": 

Add support for DNS-over-HTTPS lookups: 
 - use the "DohResolver" from the dnsjava library to make DoH lookups
 - to enable and configure it, add two new properties
    * "enableDnsOverHttpResolves" (boolean)
    * "dnsOverHttpServer" URL to the DoH Server
 - as one use case for DoH is being located behind a firewall, also support using a proxy to access the DoH server; the proxy from the FetchHTTP bean is reused in that case

Alternatively one could have introduced separate properties for the proxy to be used for DoH; if this is the preferred approach instead of making the `FetchDNS` bean factory aware to load the proxy settings from the `FetchHTTP` bean, please let me know so I can update the PR.

### DIFF
--- a/docs/configuring-jobs.rst
+++ b/docs/configuring-jobs.rst
@@ -559,16 +559,53 @@ Scripting Console
 `Heritrix3 Useful Scripts <https://github.com/internetarchive/heritrix3/wiki/Heritrix3%20Useful%20Scripts>`_ wiki page.]
 
 Configuring SOCKS5 Proxy
-~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~
 
 An optional configuration value to route Heritrix crawler traffic through a SOCKS5 proxy. This will override any set
 HTTP proxy configuration. It is facilitated by extending the `org.archive.modules.fetcher.FetchHTTP` bean with
 `socksProxyHost` and `socksProxyPort` values, as in the example below:
 
-```
-<bean class="org.archive.modules.fetcher.FetchHTTP" id="fetchHttp">
-    <!--  ... -->
-    <property name="socksProxyHost" value="127.0.0.1"/>
-    <property name="socksProxyPort" value="24000"/>
-</bean>
-```
+.. code-block:: xml
+
+    <bean class="org.archive.modules.fetcher.FetchHTTP" id="fetchHttp">
+        <!--  ... -->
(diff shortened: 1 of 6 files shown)


ato:
Nice work! Two suggestions:

1. Since it is mandatory to use both options together I think we should just have a single dnsOverHttpServer option which can be set to a URL to enable DOH. This is also how the other settings like the proxy ones work; there's no separate useHttpProxy boolean option.
2. Copying the FetchHTTP proxy settings into the global system properties seems likely to cause surprises when running multiple jobs at the same time. Since we have no option for dnsjava but to set it globally I think it might make more sense to do the reverse and have FetchHTTP default to using the values of the proxy system properties unless overridden with per-job config. We could perhaps even add a command-line option to make setting the system properties on startup more convenient:

       ./bin/heritrix --web-admin password --proxy http://localhost:8080/

    That would then enable using the proxy for all modules (FetchHTTP, FetchDNS, ...) across all jobs.

ClemensRobbenhaar:
Ah, yes, I was afraid that setting the global properties will cause trouble that I have not anticipated. For concurrent jobs this will be indeed a problem. I will try to see if I can put it the other way around as suggested.

Also ok for using a single `dnsOverHttpServer` option to enable DoH.

An update to the PR will come as soon as I finished testing the changes, which might need a little time.

ClemensRobbenhaar:
Now with an update:

* the `enableDnsOverHttpResolves` property has been scrapped in favor of using a non-empty ``dnsOverHttpServer`` to use DoH
* there are now two new command line options: ``--proxy-host`` and ``--proxy-port`` to set global proxy options. I preferred two to avoid having to parse the option value; if no ``--proxy-port`` is defined, a default value of 8000 is assumed. Also I found no intuitive short options (single key) values, so I did not define any.
* as this means that with the SOCKS5 proxy options there are now three different ways to set a proxy, I felt this warrants a separate section in the documentation. (Actually there is a fourth way by adding `` -Dhttp.proxyHost=...`` etc to the ``JAVA_OPTS``, but I did not document these.)

ato:
Looks good. Thanks!

PR #484, opened 2022-05-10
WIP: use yt-dlp instead of youtube-dl

galgeek:


adam-miller:
Closing this, since we did switch to yt-dlp already via another branch.

PR #539, opened 2022-11-16
Don't warn for non authenticating calls

krakan:
There's a warning in the `heritrix3_err.log` about failed authentication for every HTTP-request since Heritrix switched to digest authentication. That form of authentication requires an initial unauthenticated request in order to get the challenge for the subsequent authentication digest.
Also the RateLimitGuard wasn't actually limiting any authentication attempts until after many millions of failures.


ato:
Thanks!

PR #549, opened 2023-03-06
Job Status description  now shows that a job is not running anymore

fedorw:
In case of a crashed job, the crawlsummaryreport-log says: "Finished - Abnormal exit from crawl while the ApplicationContext still reports running==true. So the console shows "Job is ACTIVE - Running", while the process crashed  This change reports the exit status in this case.


anjackson:
Thanks!

PR #591, opened 2024-06-06
feat: Add logging to crawl.log for metadata records created by ExtractorYoutubeDL

adam-miller:


modules/src/main/java/org/archive/modules/warc/WARCRecordBuilder.java
```diff
@@ -43,4 +43,6 @@ public interface WARCRecordBuilder {
     WARCRecordInfo buildRecord(CrawlURI curi, URI concurrentTo)
             throws IOException;
 
+    void postWrite(WARCRecordInfo warcRecordInfo, CrawlURI curi);
+
 }
```
gretchenleighmiller, line 48: nit: missing EOF newline

gretchenleighmiller:
LGTM, but do we have a good idea of how we can test this? Would unit tests be appropriate? Can we do a full test on QA?

PR #592, opened 2024-06-08
fix: maven was not running older junit3 style tests without some extr…

adam-miller:
…a configuration.


gretchenleighmiller (approved):

ato:
Oh wow. I can't believe I didn't notice the tests were unusually pleasant by finishing in 30 seconds. :laughing:

PR #593, opened 2024-06-13
feat: Add logging to crawl.log for metadata records created by ExtractorYoutubeDL

adam-miller:


contrib/src/main/java/org/archive/modules/extractor/ExtractorYoutubeDL.java
```diff
@@ -581,6 +610,22 @@ public WARCRecordInfo buildRecord(CrawlURI curi, URI concurrentTo)
...
+        Recorder recorder = new Recorder(new File(controller.getScratchDir().getFile(), recorderBaseName),
+                controller.getRecorderOutBufferBytes(), controller.getRecorderInBufferBytes());
+        recorder.getRecordedInput().setDigest("sha1");
+        getLocalTempFile().seek(0);
+        recorder.inputWrap(Channels.newInputStream(getLocalTempFile().getChannel()));
+        recorder.getRecordedInput().startDigest();
+        recorder.outputWrap(new NullOutputStream());
+        recorder.getRecordedInput().readFully();
+        curi.getData().put(YDL_JSON_FILE_DIGEST,recorder.getRecordedInput().getDigestValue());
+        recorder.getRecordedOutput().close();
```
ato, line 626: Since we don't call recorder.cleanup() would this leak temp files when the resource is larger than the buffer size and Recorder spills to disk?

Are we only using Recorder here for the side-effect of calculating the digest? Since we have DigestUtils from Commons Codec available maybe we could just do:

```java
getLocalTempFile().seek(0);
InputStream inputStream = Channels.newInputStream(getLocalTempFile().getChannel());
curi.getData().put(YDL_JSON_FILE_DIGEST, DigestUtils.sha1(inputStream));
```

adam-miller: Yeah, my intention was to reuse existing code paths when possible, but this would simplify things, and sha1 should be pretty safe. Thanks for taking a look.

adam-miller: The hashing portion has been refactored. Much simpler now.

ato (approved):

PR #594, opened 2024-06-18
feat: AIT - Add logging to crawl.log for metadata records created by ExtractorYoutubeDL

adam-miller:


avdempsey (approved):

modules/src/main/java/org/archive/modules/warc/WARCRecordBuilder.java
```diff
@@ -43,4 +43,6 @@ public interface WARCRecordBuilder {
     WARCRecordInfo buildRecord(CrawlURI curi, URI concurrentTo)
             throws IOException;
 
-}
\ No newline at end of file
+    void postWrite(WARCRecordInfo warcRecordInfo, CrawlURI curi);
```
gretchenleighmiller, line 46: So this effectively is a new hook we have that could be generalized for similar use cases? That's pretty neat!

gretchenleighmiller (approved):

PR #595, opened 2024-07-03
Apply speculativeFixup before evaluating meta content

kris-sigur:
This avoids treating meta conent values like "Example.com" as relative urls as they are converted to absolute URLs. This is already done for speculative JS extraction.

The example sited above is common in meta "sitename" elements where the sitename is something dot com or similar.


kris-sigur:
For additional context. One of the main banks here in Iceland includes this meta tag in every page:
```
<meta property="og:site_name" content="Landsbankinn.is"/>
```

Leading to a lot of false positives. Worse, as they run a very sensitive security tool, these false positive requests are marked as malicious and the crawler is quickly blacklisted.

ato (approved):

PR #601, opened 2024-08-07
Remove ExtractorChrome

ato:
This never worked that well, is causing random test failures and I don't have any plans to continue developing it.


adam-miller (approved):

adam-miller:
I do like the idea here, and I recently did give it a try, but ran into issues. The concept is still sound, but removal does seem to be the pragmatic approach for now. Thank you.

PR #625, opened 2024-11-20
Upgrade to Spring 6.1 (and Java 17 minimum)

ato:
Spring 5 is no longer receiving open source security updates. Spring 6 requires Java 17 so we may as well set that as the new source version level.

The `@Required` annotation has been removed and they suggest using constructor injection instead. But changing to that would break existing Heritrix crawl XML files. We could just live without it but it's nice having the error message enforced at config load time rather than risking the job crashing with a more cryptic error later so I reimplemented an equivalent annotation, but marked deprecated to avoid in new beans.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
-        jdk: [8, 11, 17, 21, 23]
+        jdk: [17, 21, 23]
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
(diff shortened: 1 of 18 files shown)


ato:
1. Using StdInstantiatorStrategy instead of SerializingInstantiatorStrategy seems to be another solution. It seems to be sufficiently equivalent for our purposes.
2. The Spring 6.1 release notes [explains this and recommends](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-6.1-Release-Notes#parameter-name-retention) compiling with `-parameters` and that seems to resolve the problem without needing to use `@Qualifier`. I went with that since it would be hard to determine if there are any non-default beans relying autowiring based on parameter name.
3. Browse Beans broke again, due to Logger infinitely looping and a different exception with ConcurrentHashMap. I changed the code so it avoids recursing into Logger and treats Maps and Lists only as containers, not as beans.

PR #643, opened 2025-03-13
fix: ObjectIdentityBdbManualCacheTest.java fails with -DrunSlowTests=…

adam-miller:
…true on Java 9+

Accessibility changed in Java 9 and prevents the AtomicInteger used in this test from being accessed. The test appears to be valid, but the implementation won't work without additional JVM options. See also: https://stackoverflow.com/a/41265267

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -204,7 +204,7 @@
 					<excludes>
 						<exclude>**/Test*.java</exclude>
 					</excludes>
-					<argLine>-Xmx1g</argLine>
+					<argLine>-Xmx1g --add-opens java.base/java.util.concurrent.atomic=ALL-UNNAMED</argLine>
 				</configuration>
 			</plugin>
 		</plugins>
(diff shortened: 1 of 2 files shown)


mistydemeo:
When we looked at it we realized `AtomicInteger` is only used in the test, so it looks like the error in question wouldn't happen at runtime - just in the tests.

mistydemeo (approved):

ato:
Ah, good catch. I should probably run the slow tests before making each release.

Looks like Kryo comes with a serializer for AtomicInteger we could enable. I wonder if that might eliminate need to make the private field accessible?

e.g. in KryoBinding.java:
```java
AutoKryo kryo = new AutoKryo();
kryo.addDefaultSerializer(AtomicInteger.class, DefaultSerializers.AtomicIntegerSerializer.class);
...
```

adam-miller:
That does work, and seems like a better approach. I've updated the pull request.

PR #646, opened 2025-04-22
feat: Add serializer for Histotable used by HighestUriPrecedenceProvider

adam-miller:
This fixes an issue when using the `HighestUriPrecedenceProvider` where the nested generics in the `Histotable` fail to serialize properly. 

Sample config:
```
  <bean id="frontier" class="org.archive.crawler.frontier.BdbFrontier">
    <property name="queuePrecedencePolicy">
      <bean class="org.archive.crawler.frontier.precedence.HighestUriQueuePrecedencePolicy"/>
    </property>
    <property name="maxRetries" value="40"/>
    <property name="retryDelaySeconds" value="180"/>
    <property name="dumpPendingAtClose" value="true"/>
  </bean>
  ```

### DIFF
--- a/commons/src/main/java/org/archive/bdb/KryoBinding.java
+++ b/commons/src/main/java/org/archive/bdb/KryoBinding.java
@@ -22,9 +22,11 @@ import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 import com.esotericsoftware.kryo.serializers.DefaultSerializers;
+import com.esotericsoftware.kryo.serializers.FieldSerializer;
 import com.esotericsoftware.kryo.util.Pool;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by adam-miller** [comment]
I'm left with the following warning:
```
00:02  WARN: Class is not registered: org.archive.crawler.frontier.precedence.HighestUriQueuePrecedencePolicy$HighestUriPrecedenceProvider
Note: To register this class use: kryo.register(org.archive.crawler.frontier.precedence.HighestUriQueuePrecedencePolicy.HighestUriPrecedenceProvider.class);
```

However, this class is in the `engine` subproject and importing it into `KryoBinding` would create a circular dependency. I see that other classes such as `BdbWorkQueue` seem to get around this with an `autoregisterTo()` method, but I'm not clear on how to get this to get that function called if implemented in `HighestUriQueuePrecedencePolicy`

PR #668, opened 2025-09-18
feat: KnowledgableExtractorJS extends ConfigurableExtractorJS to inhe…

adam-miller:
…rit new config options


adam-miller:
`KnowledgableExtractorJS` has some custom code for manipulating candidate url strings discovered via JavaScript. It would be useful to extend `ConfigurableExtractorJS` instead of the base `ExtractorJS` to gain the configuration capabilities.

PR #671, opened 2025-10-04
fix: handle more bdb shutdown interrupts

adam-miller:


gretchenleighmiller (approved):

ato (approved):
Yeah, to make it robust I guess we could give each thread its own "interrupt lock". Take the lock for the duration of each BDB operation and also when calling interrupt() on it.

adam-miller:
If I understand Thread.interrupt() properly, it is just setting a flag, but that flag has special connotations for bdb. Maybe giving ToeThread its own requestStop() could replace the call to interrupt() and then switch uses of interrupted() to stopRequested(). Then maybe we wouldn't have to wrap each of the bdb interactions?

Issue #690, opened 2025-11-17
Consolidate and log on deleting URIs from BdB database


saraaubry:
In engine/src/main/java/org/archive/crawler/frontier/BdbMultipleWorkQueues.java
check if CrawlURI item and BdB holderkey are null.

PR #693, opened 2025-11-22
Fix: Add null checks and logging when deleting URIs from BdbMultipleWorkQueues (#690)

manikyarathore:
This PR fixes issue #690 by adding proper null handling and improved logging inside `BdbMultipleWorkQueues.delete()`.

Changes include:
- Added check for null CrawlURI
- Added check for null holderKey
- Type validation for holderKey to ensure it is a DatabaseEntry
- Added warning logs for all invalid states
- Prevents NullPointerExceptions during delete()

This improves stability of the frontier when corrupted or partially persisted CrawlURIs appear in the queue.


ato:
Thanks!

PR #733, opened 2026-05-15
fix: cap dynamic attribute names to avoid pathological loop.

adam-miller:
I ran into a page with an attribute value that is 250MB. We cap that in the regex, but it causes problems when searching for the next attribute and we encounter the \w* while still positioned inside the very large value.


mdellabitta (approved):
Looks good! That's a wild problem to have!

PR #760, opened 2026-08-18
Add BotBlockDetector

ato:
Detects responses produced by common bot-blocking services and adds a "botblock:service" annotation. This is intended for reporting and troubleshooting but may also be useful as a signal to stop crawling or to skip writing WARC records.

### DIFF
--- a/docs/bean-reference.rst
+++ b/docs/bean-reference.rst
@@ -521,6 +521,24 @@ ScrollDownBehavior
 
 .. bean-doc:: org.archive.modules.behaviors.ScrollDownBehavior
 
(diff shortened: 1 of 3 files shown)


nlaforwardism:
I like it - possible considerations:
* Is it worthwhile/straightforward to detect DataDome?
* re cloudflare detection, possibly worth moving the check of "cf-mitigated" (line 108) outside of the "server: cloud flare" check - their doco (as far as I could tell) seems to only mention cf-mitigated.

ato:
@nlaforwardism 👍

PR #761, opened 2026-08-21
Introduce queue-group scheduling mechanism

haja91150:
Added `QueueGroup` and `QueueGroupManager` to implement queue-grouping for shared resource management, politeness gating, and round-robin scheduling. Introduced test cases to validate functionality and integration.

The implementation adds three main pieces:
- **QueueGroup:** represents the configuration and runtime state of one group. It stores the group name, the member matchers, the maximum allowed parallelism, and the shared minimum delay. It also manages the group-level scheduling state, including the shared gate and the round-robin rotation between discovered member queues.
- **QueueGroupManager:** acts as the optional registry for all configured groups. It resolves which group, if any, a given queue belongs to, caches those lookups, and records the concrete queues discovered at runtime. It also allows groups or members to be added or removed dynamically, which can be useful from the scripting console.
- **WorkQueueFrontier integration:** the frontier remains responsible for selecting the next eligible queue, but now checks whether a queue belongs to a group before emitting its next URI. If the queue is not grouped, the original behavior is preserved. If it is grouped, the frontier applies the group gate and the group round-robin rules before allowing the URI to be fetched.

A key part of the design is the group gate.
The group gate is the shared politeness mechanism for all queues in the same group. It decides whether the group is currently allowed to emit another URI. The gate can be closed for two reasons:
- Shared concurrency is already full
	If the number of active queues in the group has reached queueGroup.getMaxParallelInGroup(), no other member of the group may start a fetch until one of the active fetches finishes.
- The group is still in shared cooldown
After a fetch finishes, the group arms a shared cooldown. The next fetch from any member of the group must wait until this cooldown expires. The effective delay is based on the normal per-URI politeness delay, but can be raised by groupMinDelayMs. _**effective delay** : max(curi.getPolitenessDelay(), groupMinDelayMs)_

When the gate is closed, the queue is not emitted. Instead, it is put aside until the group becomes eligible again, allowing the frontier to continue serving other queues outside the group, or other groups that are ready.
When the gate is open, the frontier still checks the group round-robin order. This ensures that the next URI comes from the member whose turn it is, rather than always selecting the first ready queue. Once a member is allowed through, the group acquires a shared slot. When the fetch completes, that slot is released and the group cooldown is updated.
In other words, the group gate makes the group behave like a single polite communication channel toward the target server, while the round-robin logic keeps distribution fair between the individual host queues inside that channel.

We add observability in the frontier summary report, so that configured groups expose their current state: configured members, discovered queues, active in-process count, and remaining cooldown time.
Example:
frontierSummary

We  add **queueGroup** annotation in CrawlUri to display it in Crawl Report (crawl.log).

### DIFF
--- a/engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
+++ b/engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
@@ -209,10 +209,7 @@ public class QueueGroup implements Serializable {
                 || classKey.startsWith(value + "+")) {
             return true;
         }
-        // SURT-prefix member (ends with a comma), e.g. "eu,europa,"
-        if (value.endsWith(",") && classKey.startsWith(value)) {
-            return true;
-        }
+        
         return false;
     }
 
--- a/engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
+++ b/engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
@@ -1,0 +1,416 @@
+/*
+ *  This file is part of the Heritrix web crawler (crawler.archive.org).
+ *
+ *  Licensed to the Internet Archive (IA) by one or more individual
+ *  contributors.
+ *
+ *  The IA licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.archive.crawler.frontier;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * A "queue group": a set of otherwise-independent per-host work queues
+ * that a crawler should treat as sharing a single web server for the
+ * purposes of politeness and scheduling (issue #754, "option 1").
+ *
+ * <p>The URLs are <b>not</b> merged into a single queue: each host keeps its
+ * own {@link WorkQueue} (its own {@code classKey}). The group only adds, on
+ * top of the existing flat scheduler:</p>
+ * <ul>
+ *   <li>a <b>shared politeness gate</b>: at most {@link #maxParallelInGroup}
+ *       member queues may be "in process" simultaneously, and a shared
+ *       cool-down of at least {@link #groupMinDelayMs} ms is enforced between
+ *       fetches of the group;</li>
+ *   <li>a <b>round-robin rotation</b> among the group's member queues, so the
+ *       crawler alternates between hosts (european-union.europa.eu, then
+ *       commission.europa.eu, ...) instead of draining one host first. Members
+ *       that are not currently ready (snoozed, empty) are simply skipped.</li>
+ * </ul>
+ *
+ * <p>With {@code maxParallelInGroup=1} and a shared cool-down, the whole group
+ * behaves as a single scheduling unit toward the server (the "17/50" balance
+ * described in the ticket): its member hosts share a single fetch slot.</p>
+ *
+ * <p>All rotation/gate state is transient and simply reset on
+ * restart/checkpoint.</p>
+ *
+ * <p>Members are matched against a queue's {@code classKey}. A member can be
+ * declared in three separate, dedicated lists:</p>
+ * <ul>
+ *   <li><b>{@link #groupMembersByHost}</b> (e.g. {@code european-union.europa.eu}):
+ *       exact host match;</li>
+ *   <li><b>{@link #groupMembersByRegex}</b> (e.g. {@code .*\.europa\.eu}): full
+ *       match on the host part of the classKey;</li>
+ *   <li><b>{@link #groupMembersBySurt}</b> (e.g. {@code http://(eu,europa,}):
+ *       classKey starts with the given SURT prefix.</li>
+ * </ul>
+ */
+public class QueueGroup implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private static final Logger logger =
+            Logger.getLogger(QueueGroup.class.getName());
+
+    /** Human-readable, unique name of the group (e.g. "europa_eu"). */
+    protected String name;
+
+    /** Exact host member matchers (see class javadoc). */
+    protected List<String> groupMembersByHost = new ArrayList<String>();
+
+    /** Regex member matchers, matched against the host part of the classKey. */
+    protected List<String> groupMembersByRegex = new ArrayList<String>();
+
+    /** SURT-prefix member matchers, matched against the classKey. */
+    protected List<String> groupMembersBySurt = new ArrayList<String>();
+
+    /** Max member queues collected simultaneously (shared concurrency). */
+    protected int maxParallelInGroup = 1;
+
+    /** Minimal shared delay (ms) between two fetches of the group. */
+    protected long groupMinDelayMs = 0;
+
+    // ---- transient scheduling/politeness state ----
+
+    /** Number of member queues currently "in process". */
+    protected transient int activeCount = 0;
+
+    /** Shared cool-down: no fetch before this epoch-ms. */
+    protected transient long wakeTime = 0;
+
+    /** Rotation pointer for round-robin over members. */
+    protected transient int rotationIndex = 0;
+
+    /**
+     * Concrete member queue keys discovered at runtime, in first-seen order.
+     * The configured member lists are only <i>matchers</i>; the actual
+     * per-host classKeys that fall into this group are collected here to drive
+     * the round-robin rotation. Transient: rebuilt as queues are encountered.
+     */
+    protected transient java.util.LinkedHashSet<String> discoveredMembers =
+            new java.util.LinkedHashSet<String>();
+
+    /** Cache of compiled regex patterns, keyed by the raw member string. */
+    protected transient ConcurrentHashMap<String, Pattern> regexCache =
+            new ConcurrentHashMap<String, Pattern>();
+
+    public QueueGroup() {
+    }
+
+    public QueueGroup(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public List<String> getGroupMembersByHost() {
+        return groupMembersByHost;
+    }
+
+    public void setGroupMembersByHost(List<String> groupMembersByHost) {
+        this.groupMembersByHost = (groupMembersByHost != null)
+                ? new ArrayList<String>(groupMembersByHost)
+                : new ArrayList<String>();
+    }
+
+    public List<String> getGroupMembersByRegex() {
+        return groupMembersByRegex;
+    }
+
+    public void setGroupMembersByRegex(List<String> groupMembersByRegex) {
+        this.groupMembersByRegex = (groupMembersByRegex != null)
+                ? new ArrayList<String>(groupMembersByRegex)
+                : new ArrayList<String>();
+    }
+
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
```diff
@@ -0,0 +1,416 @@
...
+            return false;
+        }
+        return classKey.startsWith(value);
+    }
+
+    /**
+     * Compile (and cache) the regex for a regex member. Returns null if the
+     * pattern is invalid (a warning is logged once per member).
+     */
+    protected Pattern regexPattern(String value) {
```
ato: I think it would be better if `groupMembersByRegex` was `List<Pattern>`, then there would be no need for this regex cache. Other configurable classes like `MatchesRegexDecideRule` just take `Pattern` directly, not `String`, I think this class should do similarly.

engine/src/test/java/org/archive/crawler/frontier/QueueGroupTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+    @Test
+    public void testSurtPrefixMatch() {
+        QueueGroup g = europa();
+        assertTrue(g.matches("http://(eu,europa,commission,"));
+        assertFalse(g.matches("http://(com,example,"));
+    }
+
+    @Test
+    public void testInvalidRegexIsIgnored() {
```
ato: Instead of silently ignoring invalid regexes it would be better to reject them up front. This happens automatically if the relevant methods take `Pattern` instead of `String`.

engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    /**
+     * @return true if the given SURT prefix is a prefix of the classKey.
+     */
+    protected boolean matchesSurt(String value, String classKey) {
+        if (value == null || value.isEmpty() || classKey == null) {
+            return false;
+        }
+        return classKey.startsWith(value);
```
ato, line 234: What's the difference between `matchesSurt` and the "SURT-prefix member" line in `matchesHost()`. Do we need both?

engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
```diff
@@ -0,0 +1,416 @@
...
+    protected boolean matchesHost(String value, String classKey) {
+        if (value == null || value.isEmpty() || classKey == null) {
+            return false;
+        }
+        if (classKey.equals(value)) {
+            return true;
+        }
+        // hostname policy suffixes: host#port and host+subqueue
+        if (classKey.startsWith(value + "#")
+                || classKey.startsWith(value + "+")) {
```
ato, line 209: The example in `profile-crawler-beans.cxml` implies this is expected to take a normal host (commission.europa.eu) instead of a surt (eu,europa,commission,). I don't think that this will work with the default SurtAuthorityQueueAssignmentPolicy. Won't classKey be in SURT form?

engine/src/main/java/org/archive/crawler/frontier/QueueGroupManager.java
```diff
@@ -0,0 +1,163 @@
...
+                removed = true;
+            }
+        }
+        if (removed) {
+            clearCache();
+        }
+        return removed;
+    }
+
+    public synchronized boolean addMember(String groupName, String member) {
```
ato, line 128: This method seems unnecessary.

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
```diff
@@ -91,31 +92,82 @@ public abstract class WorkQueueFrontier extends AbstractFrontier
...
+    /**
+     * Release the shared group slot held by the given queue and arm the shared
+     * cool-down. No-op when the queue is not grouped or there is no manager.
+     *
+     * @param wq       the finished queue
+     * @param now      current epoch-ms
+     * @param delay_ms the per-URI politeness delay just computed
+     */
+    protected void releaseGroup(WorkQueue wq, long now, long delay_ms) {
+        QueueGroup group = groupFor(wq);
```
ato, line 165: Is it possible for the group assignment to be changed or removed while a URI is processing? If so this would release the wrong group which would leave the original group broken due to `activeCount` never being decremented.

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
```diff
@@ -652,89 +704,123 @@ && highestPrecedenceWaiting < getPrecedenceFloor()) {
...
+                        String memberKey = readyQ.getClassKey();
+                        long now = System.currentTimeMillis();
+                        synchronized (group) {
+                            // (a) shared gate closed: at concurrency ceiling
+                            // or in cool-down. Park the queue (snooze until
+                            // the group re-opens) and serve another queue.
+                            if (!group.canProceed(now)) {
+                                inProcessQueues.remove(readyQ);
+                                long until = Math.max(group.getWakeTime(), now + 1);
+                                snoozeQueue(readyQ, now, until - now);
```
ato: Does this mean queues that are blocked by the concurrency ceiling will cycle between snoozed and ready up to every 1ms? This might not be too much overhead for smaller groups, but might become a problem for larger ones. Something to be careful of at least.

An alternative might be to have each QueueGroup have a waiting list to park these on and pop a queue from the waiting into ready after each release.

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
```diff
@@ -652,89 +704,123 @@ && highestPrecedenceWaiting < getPrecedenceFloor()) {
...
+                                readyQ.makeDirty();
+                                readyQ = null;
+                                continue findauri;
+                            }
+                            // (b) not this member's turn: another member is
+                            // ahead in the rotation and ready. Requeue and
+                            // retry so the expected member gets served next.
+                            if (!group.isTurn(this::isMemberReady, memberKey)) {
```
ato: This `isTurn` mechanism seems a bit strange: aren't the queues already globally approximately round-robin?

This presumably also doesn't handle precedence. If that's intended we should at least document different precedence queues in the same group isn't supported.

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
```diff
@@ -952,9 +1038,10 @@ protected void processFinish(CrawlURI curi) {
                 }
                 long delay_ms = retryDelayFor(curi) * 1000;
                 curi.processingCleanup(); // lose state that shouldn't burden
-                                          // retry
+                // retry
                 wq.unpeek(curi);
                 wq.update(this, curi); // rewrite any changes
+                releaseGroup(wq, now, delay_ms);
```
ato, line 1044: Do we actually want to apply the retry delay to the group? This means a long retry for one host will block all hosts in the group.

ato (changes requested):
Can you please remove the unrelated whitespace changes in WorkQueue and WorkQueueFrontier? It makes the changes difficult to review and if merged will make `git blame` unhelpful.

engine/src/main/resources/org/archive/crawler/restlet/profile-crawler-beans.cxml
```diff
@@ -469,7 +469,45 @@ http://example.example/example
   <!-- <property name="dumpPendingAtClose" value="false" /> -->
  </bean>
  
- <!-- URI UNIQ FILTER: Used by frontier to remember already-included URIs --> 
+ <!-- QUEUE GROUPS (issue #754): OPTIONAL, opt-in. When this bean is absent
+      (as by default, being commented out below), the frontier behaves
+      exactly as before. Uncomment and adapt to group several per-host queues
+      so they share a single politeness gate (shared concurrency + shared
+      cool-down) and are served round-robin, like a single visitor toward one
+      web server. See docs/QUEUE_GROUPS.md. -->
```
ato: 1. There is no `docs/QUEUE_GROUPS.md` in this PR.
2. The default profile shouldn't reference a GitHub issue.
3. "behaves exactly as before" doesn't make sense in this context: the bean is already commented out.

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
haja91150: Yes the queue group configuration is mutable at runtime through the scripting-console CRUD methods so that’s a valid concern. The group should be released based on the group that was actually acquired, not based on the current group assignment at finish time. I’ll adjust the implementation to store the acquired group for the in-progress URI/queue and release that exact group, which should prevent activeCount from getting stuck.

engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
haja91150: Sorry my IDE had reformatting the code i think. I will correct that and push the update.

engine/src/test/java/org/archive/crawler/frontier/QueueGroupTest.java
haja91150: Ok i will modify to Pattern.

engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
haja91150: Ok i will modify to Pattern.

haja91150:
Hello Alex, thanks for your reviews. I'll look at your comments and modify the code accordingly. I'm actually not available this week but i'll do the change next week.

engine/src/main/java/org/archive/crawler/frontier/QueueGroup.java
haja91150: You are right, we don’t need both. matchesSurt() is the intended implementation for entries configured via groupMembersBySurt. The check in matchesHost() only allowed a SURT prefix to be accidentally placed in the host list, and we shouldn't permit that implicit misconfiguration. I’ll remove the unnecessary fallback from matchesHost().

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
haja91150: **readyClassQueues** already gives us a global round-robin: it's a FIFO of ready queues, so ready queues are served roughly in turn. However, that ordering is group-agnostic and only fair on average. We (at BnF) would like to have the possibility of fairness inside a group.

The problem **isTurn** addresses is starvation inside a group. When the shared gate opens, without **isTurn** the member served is simply whichever one happens to be at the head of the global FIFO at that instant. A very prolific host in the group (one that continuously discovers new URIs and is therefore almost always "ready") can keep grabbing the shared slot and starve the slower members of the same group — server politeness stays correct, but the distribution between hosts of the group becomes unbalanced.

**isTurn + advanceRotation** add a deterministic, intra-group fairness: they keep their own rotation pointer over the concrete members actually discovered, and skip members that aren't ready (snoozed/empty), so a silent host never blocks the group while a prolific one can't monopolize it either.

If it’s ok for you, i’ll make its behavior configurable per group (a **roundRobinEnabled** flag on QueueGroup). When disabled, only the shared politeness gate applies and ordering falls back to the frontier's global round-robin; when enabled, we keep the deterministic intra-group fairness. That way each group can pick "pure politeness" or "strict fairness" as needed, and i'll add the precedence limitation with this group logic to the docs.

ato: My intuition keeps telling me that if order is preserved while blocked on the gate starvation shouldn't happen. I'm having trouble reasoning about what fairness means in your example with one queue that's often exhausted though.

I think I'll just acknowledge that you've obviously spent more time thinking about this than I have and accept the intra-group round-robin as necessary. You don't need to add an option just for my sake. :-)

haja91150: Ok, i'll keep it and add the precedence logic documented. I have made a doc in "docs/QUEUE_GROUP.md" but is there another place i should put it ?

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
haja91150: You are right, i'll change delay_ms to 0 to not block all host

engine/src/main/java/org/archive/crawler/frontier/WorkQueueFrontier.java
ato: As this is user documentation, a new section in the configuration guide seems appropriate: docs/configuring-jobs.rst

Issue #748, opened 2026-07-07
New or extension of current SourceTags report


saraaubry:
Current version of Heritrix enables to create a SourceTags report from the crawllog: source-report.txt. This report contains a line item for each host, which includes the seed from which the host was reached and the number of URLs crawled for this seed host combination..

Sample output:
```
[source] [host] [#urls]
https://www.nouvelobs.com/rue89/ focus.nouvelobs.com 6868
https://www.nouvelobs.com/rue89/ fastly.a9fast.com 287
https://www.nouvelobs.com/rue89/ www.nouvelobs.com 274
https://www.nouvelobs.com/rue89/ boutique.nouvelobs.com 155
https://www.nouvelobs.com/rue89/ cours-francais.nouvelobs.com 132
https://www.nouvelobs.com/rue89/ jeux.nouvelobs.com 130
https://www.nouvelobs.com/rue89/ img.nealis.fr 68
https://www.nouvelobs.com/rue89/ dns: 59
```

To get a better view on possible issues (in particular blocking issues with 426 429 5XX -2), we propose to add the distribution of HTTP response code and Heritrix status among the host+seed.

Sample output:
```
[source] [host] [rescode] [#urls]
https://www.sudouest.fr live.sportall.tv 200 2
https://www.sudouest.fr maps.googleapis.com 200 1
https://www.sudouest.fr media.sudouest.fr 403 9143
https://www.sudouest.fr module.hellocv.com 403 1
https://www.sudouest.fr monorail-edge-ca.shopifycloud.com -6 1
https://www.sudouest.fr monorail-edge-staging.shopifycloud.com 405 1
https://www.sudouest.fr monorail-edge.shopifysvc.com 405 1
https://www.sudouest.fr newsletter.sudouest.fr 403 1  
```

PR #766, opened 2026-09-08
Include Heritrix status code in extended SourceTags report

bnfleb:
Related to issue #748

The new extended version of the SourceTags report contains also the Heritrix status codes. This change doesn't apply to the default version of the SourceTags report.

### DIFF
--- a/engine/src/main/java/org/archive/crawler/reporting/StatisticsTracker.java
+++ b/engine/src/main/java/org/archive/crawler/reporting/StatisticsTracker.java
@@ -622,7 +622,7 @@ public class StatisticsTracker
      */
     protected static void incrementMapCount(ConcurrentMap<String,AtomicLong> map, 
             String key) {
-    	incrementMapCount(map,key,1);
+        incrementMapCount(map,key,1);
     }
     
     /**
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/archive/crawler/reporting/StatisticsTracker.java
```diff
@@ -839,22 +831,47 @@ protected void tallySourceStats(CrawlURI curi) {
...
+            }
+            if (isCrawlOK) {
+                saveSourceStats(curi.getSourceTag(), sc.getHostFor(curi.getUURI()).getHostName());
+                tally = true;
+            }
+            if (tally) {
+                tallySourceStats(curi);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ato: I'm not worried about getReport() returning null.

The second part of this looks like a real bug but I think it should be fixed in SourceTagsReport not here.

in SourceTagsReport.java:
```java
// this should probably be sourceHostStatusDistribution if isIncludeResCode
Set<String> sourceTags = stats.sourceHostDistribution.keySet();

if(sourceTags.isEmpty()) {
    writer.println("No source tag information. (Is 'sourceTagSeeds' enabled?)");
    return; 
}
```

engine/src/main/java/org/archive/crawler/reporting/StatisticsTracker.java
```diff
@@ -839,22 +831,47 @@ protected void tallySourceStats(CrawlURI curi) {
...
     }
     
+    protected void manageSourceStats(CrawlURI curi, boolean isCrawlOK) {
+        ServerCache sc = serverCache;
+        if (getTrackSources() && curi.getData().containsKey(A_SOURCE_TAG)) {
+            boolean tally = false;
+            SourceTagsReport str = getReport(SourceTagsReport.class);
+            if (str.isIncludeResCode()) {
+                saveSourceStats(curi.getSourceTag(), sc.getHostFor(curi.getUURI()).getHostName(), String.valueOf(curi.getFetchStatus()));
+                tally = true;
```
ato: This means we call `tallySourceState` for disregarded/failed/retried URIs and increment the `novelCount` tally in `CrawledBytesHistotable.accumulate()`. That seem surprising and probably incorrect? For example if the user set a SourceQuotaEnforcer on `novelCount` I don't think they'd expect enabling `includeResCode` would cause the enforced quota to suddenly count retries and failures.

ato (changes requested):
@bnfleb Can you check the check the comment about `includeResCode` affecting `novelCount`? If you think that's actually correct let me know and I'll merge this, but it seems odd.

engine/src/main/java/org/archive/crawler/reporting/StatisticsTracker.java
bnfleb: You're right. The call to `tallySourceState` must be done only if the URI has been effectively done, so when `isCrawlOK = true`.

PR #770, opened 2026-09-15
FetchHTTP: digest chunked responses after transfer decoding

### LINKED ISSUE #769: Incorect WARC-Payload-Digest with chunked encoding
When `FetchHTTP` retrieves an HTTP/1.1 response using `Transfer-Encoding: chunked`, the content digest appears to be calculated over the raw chunked message body rather than the transfer-decoded payload. This makes the digest depend on chunk boundaries. Two responses containing identical payload bytes can therefore receive different `WARC-Payload-Digest` values, preventing deduplication and causing validating errors.

This will not usually occur with the default Heritrix configuration as the HTTP standard requires servers to not send chunked responses to a HTTP/1.0 request. So normally this will only be seen if `useHTTP11` is enabled.

This does not affect `FetchHTTP2` in HTTP/1.1 mode as it decodes the transfer-encoding before recording.

Reported-by: @thomasegense

ato:
Recalculate content digests from the entity replay stream for chunked responses so that chunk framing is not included in WARC-Payload-Digest.

Fixes #769


larsnaesbye:
It would be greatly appreciated if this could get into next release. At KB we would benefit a lot from this fix.

vilmibm:
Does this interact with the warc spec at all? I skimmed it and didn't see anything but might have missed a relevant section. Just thinking ahead to whether any validation in our toolchain might not like this.

ato:
Yes, this also corrects it with respect to the WARC spec. The correct interpretation of payload in the spec was discussed and clarified in https://github.com/iipc/warc-specifications/issues/22

> two responses with different transfer-encodings (perhaps even because of different 'chunked' boundaries), but which (when the transfer-encoding is 'removed') result in the exact same 'entity-body' (per HTTP), should have the same WARC-Payload-Digest.

